PR #1, opened 2019-05-11
In memory feed store implementation

mihaicris:
Hi, this is my first attempt. I am not sure if there is a need for a singleton for this kind of simple memory store.  For this implementation I couldn't the find to use also the tests for failable operations. 
Of course Core Data has also an in memory persistence store, but I plan to try also a Core Data implementation with the classic SQLite store instead. 
Thanks for this nice idea as exercise!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+DerivedData/
+xcuserdata/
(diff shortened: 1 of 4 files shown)


caiozullo:
Hi Mihai, congratulations you've passed the challenge! 🥇

For this challenge, you don't need a singleton. Also, the `InMemoryFeedStore` doesn't need the failable operations tests since it will never fail. 

Great job! ✅ 

We're looking forward to your `CoreData` implementation. ⛩

PR #2, opened 2019-05-18
Here is my implementation of CodableFeedStore

achirkof:
Hi! Here is my implementation of CodableFeedStore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+DerivedData/
+xcuserdata/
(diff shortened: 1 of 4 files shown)


caiozullo:
Hi @achirkof, congratulations you've passed the challenge! 🥇

The tests are passing, so your implementation complies to the specs. However, you're not persisting the *timestamp* to disk (you store it as an in-memory property `var timestamp: Date?`). This means your implementation will only work as an *In-Memory* store since you cannot retrieve the persisted data from disk if the in-memory `timestamp` is `nil`.

See [line 91](https://github.com/essentialdevelopercom/ios-lead-essentials-feed-store-challenge/pull/2/files#diff-5d6b680f6ced8fa3d25bfd014ec98d4aR91):

```
if let data = self.fileManager.contents(atPath: self.path), let time = self.timestamp {
```

If you want to be able to retrieve the persisted data across instances, you need to store the `timestamp` to disk too, instead of storing it in an in-memory property. 🤝⛩

PR #3, opened 2019-09-22
FeedStore implementation as UserDefaultsFeedStore

sirEven:
Hi Caio and Mike,

This is my solution for the FeedStore Challenge. 
I have one specific question concerning the dependency diagram: As you can see, I also added ‚LocalFeedImage‘ at the top left, as my Type ‚LocalFeed‘ has a reference to it. However, there are also dependencies towards ‚LocalFeedImage‘ coming from my mapping (coding / encoding) logic in extensions to ‚Feed‘ and ‚FeedImage‘ which can be found at the bottom of the diagram. These are two private extensions with initialisers that hold the mentioned mapping logic. Should such a dependency be reflected by the diagram as well? 

Thanks a lot for your time and your insanely awesome content!

Cheers, 
Severin


caiozullo:
Hello @sirEven, congratulations you've passed the challenge! 🥇

All tests are passing on the CI server, so your implementation complies to the specs. ✅ 

Good job with the diagram too! 

Here are some minor suggestions:
- You could have avoided the `try!` when encoding/decoding the data and handle errors instead.
- We noticed you had issues migrating the project to Swift 5. To do so, you need to update the CI config to use a [macOS image](https://docs.travis-ci.com/user/reference/osx/#macos-version) that supports the Swift version you want (e.g., osx_image: xcode11).

Keep up the great work. ⛩

sirEven:
Hi @caiozullo, thanks for your helpful feedback! I implemented the 'FailableRetrieveFeedStoreSpecs' to handle the errors that those try blocks would throw. 

Cheers, 
Severin

caiozullo:
Nice one! ✅

tomasreyes (approved):

PR #5, opened 2019-11-23
Codable, Realm and InMemory implementations

danillahtin:
Hi Guys!
In addition to all the implementations I created persistent integration tests to test my implementations in integration.


Tests/PersistentFeedStoreIntegrationTests.swift
```diff
@@ -0,0 +1,206 @@
...
+import RealmSwift
+
+
+private protocol FeedStoreImplementation {
+    func setupCleanState()
+    func build() -> FeedStore
+}
+
+
+private struct RealmImplementation: FeedStoreImplementation {
```
caiozullo: I notice this implementation is never used, so the Realm stack is not tested in integration.

danillahtin: Yes, this was kid of a problem. I have two persistent implementations (realm and codable). So to test two persistent implementations I should write two test cases. But they should be identical since integration tests should not depend on implementation details otherwise it would more like isolation testing. I thought of subclassing one test case to reuse it, but in this case it would be hard to understand which of the implementations is failing. So I decided to write one test case and inject the exact implementation on demand since I think in real world project there will be no two implementations of the same functionality cause we just don't need it. So to test RealmFeedStore in integration you have to just replace CodableImplementation with RealmImplementation and just run the tests. Or it would be better to just duplicate tests for each implementations?

caiozullo: Instead of subclassing or duplicating, you could follow a similar approach as the `FeedStoreSpec` protocols: 

1. Create a spec protocol with reusable helper extensions. For example, `PersistentFeedStoreSpec`.
2. Create two separate test case classes implementing the spec protocol and reusing the helpers.

This way, you can test your implementations in isolation, and you don't need any manual steps to run the full test suite. ✅

PR #9, opened 2020-02-01
Challenge feed store virych

damienissa:


FeedStoreChallenge/FeedStore/FileSystemFeedStore.swift
```diff
@@ -0,0 +1,97 @@
...
+        set {
+            save(newValue)
+        }
+    }
+    
+    public init() { }
+    
+    private func save(_ cache: Cache?) {
+        
+        if cache == nil {
```
caiozullo: If you use a `guard let` here, you won't have to force-unwrap the optional below (`cache!`).

FeedStoreChallenge/FeedStore/Realm/RealmFeedStore.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+//  Created by Dima Virych on 01.02.2020.
+//  Copyright © 2020 Essential Developer. All rights reserved.
+//
+
+import Foundation
+
+public class RealmFeedStore {
+    
+    var realm: RealmDecorator!
```
caiozullo: Since you're using Constructor Injection, this variable doesn't the implicitly unwrapped optional (!).

It could be 

`var realm: RealmDecorator`

or 

`let realm: RealmDecorator`

Tests/FileSystemFeedStoreChallengeTests.swift
```diff
@@ -0,0 +1,105 @@
...
+//  Tests
+//
+//  Created by Dima Virych on 31.01.2020.
+//  Copyright © 2020 Essential Developer. All rights reserved.
+//
+
+import XCTest
+import FeedStoreChallenge
+
+fileprivate let dbURL = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!.appendingPathComponent("\(FileSystemFeedStore.self).db")
```
caiozullo: You could inject a test-specific URL into the `FileSystemFeedStore` store, so the test wouldn't need to know about the exact production URL (which is an implementation detail).

PR #20, opened 2020-05-22
In memory cache implementation

panditpakhurde:
I have implemented InMemoryFeedStore with NSCache as a persistent stack. 
Please verify and do let me know your views and best possible way to implement the same.


FeedStoreChallenge/InMemoryCacheFeedStore.swift
```diff
@@ -0,0 +1,57 @@
...
+    
+    public init(key: String) {
+        self.key = Key(key)
+    }
+    
+    public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+        if let _ = memoryCache.object(forKey: key) {
+            memoryCache.removeObject(forKey: key)
+        }
+
```
caiozullo, line 41: You don't need to check if the value exists:

```suggestion
        memoryCache.removeObject(forKey: key)
```

FeedStoreChallenge/InMemoryCacheFeedStore.swift
```diff
@@ -0,0 +1,57 @@
...
+            self.key = key
+        }
+    }
+    
+    private var key: Key
+    private var memoryCache = NSCache<Key, Cache>()
+    
+    public init(key: String) {
+        self.key = Key(key)
+    }
```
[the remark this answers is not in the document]
devessentials20: > It doesn't look like you need your own `Key` type and inject the value here. You could make the key private and use an `NSString` or `NSUUID`, for example:

Yeah.. it make sense. I overthink on this. I though to make feedstore instance to be used to store more than one feedstore. But injecting key with initialiser convert it into to have one feed store only. Thank you for the review.

FeedStoreChallenge/InMemoryCacheFeedStore.swift
devessentials20: Yeah, right.

devessentials20:
> Hi @devessentials20,
> 
> Congratulations, you've passed the challenge! 🥇
> 
> * All tests are passing, so your implementation complies with the specs.
> 
> Great job ✅
> 
> I understand you chose `Array` and `NSCache` for your solutions as an exercise. But I want to point out that you don't need to hold the value in a collection for this challenge.
> 
> Since you're managing just one value at a time, `Optional` would do just fine for an in-memory implementation:
> 
> ```
> private var cache: (feed: [LocalFeedImage], timestamp: Date)?
> ```
> 
> Moreover, I'm not sure the benefits of injecting the `Key` since the `NSCache` is a private implementation detail. I added a suggestion to keep the `Key` private.
> 
> Let me know what you think - thanks!



> Hi @devessentials20,
> 
> Congratulations, you've passed the challenge! 🥇
> 
> * All tests are passing, so your implementation complies with the specs.
> 
> Great job ✅
> 
> I understand you chose `Array` and `NSCache` for your solutions as an exercise. But I want to point out that you don't need to hold the value in a collection for this challenge.
> 
> Since you're managing just one value at a time, `Optional` would do just fine for an in-memory implementation:
> 
> ```
> private var cache: (feed: [LocalFeedImage], timestamp: Date)?
> ```
> 
> Moreover, I'm not sure the benefits of injecting the `Key` since the `NSCache` is a private implementation detail. I added a suggestion to keep the `Key` private.
> 
> Let me know what you think - thanks!



> ```
> private var cache: (feed: [LocalFeedImage], timestamp: Date)?
> ```

Thanks you for detailed review. I have done changes in the InMemoryFeedStore and InMemoryCacheFeeStore as per your suggestion and updated the pull requests. Please check.

caiozullo:
Great job 🙂

devessentials20:
> Great job 🙂

Thank you 🙂

PR #23, opened 2020-06-05
CoreData implementation

mduranx64:


FeedStoreChallenge/Feed Cache/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,143 @@
...
+    }()
+    
+    private var context: NSManagedObjectContext {
+        return persistentContainer.viewContext
+    }
+    
+    public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+        do {
+            try deleteCache()
+            completion(nil)
```
caiozullo, line 106: Missing `try context.save()`

Tests/FeedStoreChallengeTests.swift
```diff
@@ -12,136 +12,112 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+    private func undoStoreSideEffects() {
+        deleteStoreDataBase()
+    }
+    
+    private func deleteStoreDataBase() {
+        try? FileManager.default.removeItem(at: testSpecificStoreURL())
+    }
+    
+    private func testSpecificStoreURL() -> URL {
+        return cachesDirectory().appendingPathComponent("database")
```
caiozullo, line 117: You could use `/dev/null` to speed up the tests and avoid writing side-effects to disk.

caiozullo, line 117: You could use `/dev/null` to speed up the tests and avoid writing artifacts to disk.

FeedStoreChallenge/Feed Cache/CoreDataFeedStore.swift
caiozullo, line 106: Missing `try context.save()` - you could check this behavior with an integration test.

FeedStoreChallenge/Feed Cache/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,143 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Miguel Duran on 03-06-20.
+//  Copyright © 2020 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+public class CoreDataFeedStore: FeedStore {
```
caiozullo, line 12: At the moment the implementation works fine on the main thread. To make your implementation thread-safe, you can wrap the insert/retrieve/delete operations in a `context.perform` block:

https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506578-perform

FeedStoreChallenge/Feed Cache/CoreDataFeedStore.swift
mduranx64: I added an integration test and it fail when `try context.save()` is missing.

Tests/FeedStoreChallengeTests.swift
mduranx64: I change the tests to use `/dev/null`.

FeedStoreChallenge/Feed Cache/CoreDataFeedStore.swift
mduranx64: Thank you for the detailed review
I change the implementation to use `context.perfom` block

caiozullo:
Nice one!

PR #24, opened 2020-06-22
`InMemoryFeedStore` - Iván Camilo Fuertes

iCamilo:
see `README -  The Dependency Diagram demonstrating the architecture of your solution` for more information about this PR.


Tests/FeedStoreChallengeTests.swift
```diff
@@ -7,141 +7,121 @@ import FeedStoreChallenge
...
+        trackForMemoryLeak(sut)
+        
+        return sut
 	}
+    
+    private func trackForMemoryLeak(_ instance: FeedStore, file: StaticString = #file, line: UInt = #line) {
+        var weakInstance: FeedStore? = instance
+        weakInstance = nil
+        
+        XCTAssertNil(weakInstance, "Instance should be correctly deallocated to avoid memory leaks", file: file, line: line)
```
caiozullo, line 122: This implementation is not checking for potential memory leaks. You need to check `XCTAssertNil` in a `tearDown` block holding a weak reference to the instance.

Example: https://github.com/essentialdevelopercom/essential-feed-case-study/blob/master/EssentialFeed/EssentialFeedTests/Helpers/XCTestCase%2BMemoryLeakTracking.swift

iCamilo: * And I couldn't replicate the failing tests,

Thanks for checking 🖖. Here I keep seeing failing test every now and then but I do not understand what is going on. 

* I left just a comment regarding the trackForMemoryLeak helper.
   Done and Thanks ! 

I am still having some issues understanding what causes a memory leak or not
For me, this should cause a retain cycle but according to the tests it does not 


```
FeedStoreChallengeTests

let sut = InMemoryFeedStore()
```

```
InMemoryFeedStore

private let queue 

  queue.async(flags: .barrier) {             
            self.cache = nil 
  }
}
```

`InMemoryFeedStore` has a strong reference to `queue` and `queue` `async` captures self. I see there a retain cycle alright, but the tests say that is not the case 🙌.
There is something that I am missing even after rereading Swift documentation for closures. 

`let sut = InMemoryFeedStore()` ARC +1
`queue.async(flags: .barrier) {   self.cache = nil  }` ARC +2 

deallocating `sut`, I understand decrements ARC in one, so there should be one reference lingering somewhere, but it seems is not the case. 

Could you please explain to me why this is not a retain cycle? 

Thanks!

caiozullo: Hi @iCamilo. There's no retain cycle because the `queue` doesn't hold a strong reference to the closure after execution.

PR #32, opened 2020-09-21
CoreData Implementation - Nicolas De Maio

Nico113:


FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,99 @@
...
+        self.bundle = bundle
+    }
+    
+    lazy var persistentContainer: NSPersistentContainer? = {
+        if let mom = NSManagedObjectModel.mergedModel(from: [bundle]) {
+            let container = NSPersistentContainer(name: modelName, managedObjectModel: mom)
+                    
+            container.loadPersistentStores(completionHandler: { (storeDescription, error) in
+                if let error = error as NSError? {
+                    fatalError("Unresolved error \(error), \(error.userInfo)")
```
caiozullo, line 27: It'd be best to handle the possible error. For example, you could throw the error, so the clients know something went wrong. 

Moreover, you shouldn't be able to instantiate a `CoreDataFeedStore` with a wrong model name or bundle. So you could run those checks in the initializer. If it fails to find the `NSManagedObjectModel` or create the `NSPersistentContainer`, it should throw.

So you never have a `CoreDataFeedStore` instance that doesn't work (e.g., it can't work without an NSPersistentContainer!)

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,99 @@
...
+public class CoreDataFeedStore: FeedStore {
+    private let modelName: String
+    private let bundle: Bundle
+    
+    public init(model: String = "FeedStore", bundle: Bundle = Bundle(for: CoreDataFeedStore.self)) {
+        modelName = model
+        self.bundle = bundle
+    }
+    
+    lazy var persistentContainer: NSPersistentContainer? = {
```
caiozullo, line 21: This property can be private - it's an implementation detail.

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,99 @@
...
+                    fatalError("Unresolved error \(error), \(error.userInfo)")
+                }
+            })
+            
+            return container
+        }
+        return nil
+    }()
+    
+    func deleteCurrentCache() throws {
```
caiozullo, line 36: This method can be private - it's an implementation detail.

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,99 @@
...
+        if let container = persistentContainer, let currentCache: Cache = try container.viewContext.fetch(Cache.fetchRequest()).first {
+            container.viewContext.delete(currentCache)
+            try container.viewContext.save()
+        }
```
caiozullo, line 40: Every CoreData operation must run within the `NSManagedObjectContext` queue. You can do so with:

```
context.perform {
  // run operation here (query/delete/save)
}
```

The current code is only working because the tests and the `viewContext` both run in the main queue. But prod code may invoke this method from other queues.

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,99 @@
...
+    public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+        do {
+            try deleteCurrentCache()
+            completion(.none)
+        } catch {
+            completion(error)
+        }
+    }
+    
+    public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
```
caiozullo, line 52: Same here, the operations should be wrapped in a `context.perform {}` closure.

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,99 @@
...
+                completion(.none)
+            } catch {
+                completion(error)
+            }
+        } else {
+            completion(NSError())
+        }
+    }
+    
+    public func retrieve(completion: @escaping RetrievalCompletion) {
```
caiozullo, line 84: Same here, the operations should be wrapped in a `context.perform {}` closure.

FeedStoreChallenge/CoreData Feed Store/Cache+CoreDataProperties.swift
```diff
@@ -0,0 +1,61 @@
...
+import Foundation
+import CoreData
+
+extension Cache {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<Cache> {
+        return NSFetchRequest<Cache>(entityName: "Cache")
+    }
+
+    @NSManaged public var timestamp: Date?
```
caiozullo, line 19: `timestamp` should be non-optional

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,99 @@
...
+                var feedImages = [NSManagedObject]()
+                for localFeedImage in feed {
+                    let feedImage = NSManagedObject(entity: feedImageEntity, insertInto: container.viewContext)
+                    
+                    feedImage.setValue(localFeedImage.id, forKey: "id")
+                    feedImage.setValue(localFeedImage.description, forKey: "imageDescription")
+                    feedImage.setValue(localFeedImage.location, forKey: "location")
+                    feedImage.setValue(localFeedImage.url, forKey: "url")
+                    
+                    feedImages.append(feedImage)
```
caiozullo, line 68: You can use the `FeedImage` class here, which is a subclass of `NSManagedObject`. So you can use the setters:

```
let feedImage = FeedImage(...)
feedImage.id = localFeedImage.id
feedImage. imageDescription = localFeedImage.description
...
```

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,99 @@
...
+                }
+                
+                cache.timestamp = timestamp
+                cache.addToItems(NSOrderedSet(array: feedImages))
+                
+                try container.viewContext.save()
```
caiozullo, line 74: You could perform every operation in a background context - so you don't perform expensive operations in the main queue (one of the benefits is that the UI can be more responsible, as it runs in the main queue).

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,99 @@
...
+        } else {
+            completion(NSError())
+        }
+    }
+    
+    public func retrieve(completion: @escaping RetrievalCompletion) {
+        if let container = persistentContainer {
+            do {
+                if let cache: Cache = try container.viewContext.fetch(Cache.fetchRequest()).first {
+                    completion(.found(feed: cache.local, timestamp: cache.timestamp!))
```
caiozullo, line 88: If you make the `timestamp` non-optional, you don't need to force-unwrap it here.

FeedStoreChallenge/CoreData Feed Store/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="17192" systemVersion="19G2021" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="Cache" representedClassName="Cache" syncable="YES">
+        <attribute name="timestamp" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
```
caiozullo, line 4: `timestamp` should be non-optional

Tests/FeedStoreChallengeTests.swift
```diff
@@ -7,150 +7,106 @@ import FeedStoreChallenge
...
+    }
+    
+    private func clearCache(file: StaticString = #file, line: UInt = #line) {
+        let sut = makeSUT()
+        
+        sut.deleteCachedFeed { (error) in
+            if let error = error {
+                fatalError("Unresolved error \(error)")
+            }
+        }
```
caiozullo, line 109: This call is async, so the `clearCache` method can return before the cache deletion completes. It's important that you wait for the operation to complete, otherwise, it can affect the result of other tests.

You could use the test helper that wait for the cache deletion with a timeout:

```
deleteCache(from: sut)
```

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -31,60 +31,71 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+    }
+    
+    private func clearCache(file: StaticString = #file, line: UInt = #line) {
+        let sut = makeSUT()
+        
+        sut.deleteCachedFeed { (error) in
+            if let error = error {
+                fatalError("Unresolved error \(error)")
+            }
+        }
```
caiozullo, line 98: This call is async, so the `clearCache` method can return before the cache deletion completes. It's important that you wait for the operation to complete, otherwise, it can affect the result of other tests.

You could use the test helper that wait for the cache deletion with a timeout:

```
deleteCache(from: sut)
```

Tests/FeedStoreChallengeTests.swift
```diff
@@ -7,150 +7,106 @@ import FeedStoreChallenge
...
-    //
-    //  Repeat this process until all tests are passing.
-    //
-    //  ***********************
-
+    override func setUp() {
+        super.setUp()
+        
+        clearCache()
+    }
```
caiozullo, line 14: If your implementation leaves side-effects, it's important that you also clean it in `tearDown`. Every test should leave the system in a clean state.

caiozullo (commented):
Hi @Nico113!

Your implementation looks great, and all tests are passing. But to pass the challenge, you need to address some issues, such as handling errors and running operations in the context queue with `context.perform {}`.

I left some comments to guide you. Feel free to push your changes and let me know when you're done ✅

Nico113:
Hey @caiozullo I think I have addressed all the comments you made, thanks for your feedback.
Btw I had to leave trackForMemoryLeak method call commented because I am getting the following error: "API violation -- attempting to add a teardown block after teardown blocks have been dequeued (NSGenericException)" and I couldn't figure it out how to solve it.
I know why this is happening, due to I am cleaning the artifacts in the tear down method and also I am adding another block of code that is being executed before the tear down but I am not sure how to fix it. Any ideas?

caiozullo:
> Btw I had to leave trackForMemoryLeak method call commented because I am getting the following error: "API violation -- attempting to add a teardown block after teardown blocks have been dequeued (NSGenericException)" and I couldn't figure it out how to solve it.

> I know why this is happening, due to I am cleaning the artifacts in the tear down method and also I am adding another block of code that is being executed before the tear down but I am not sure how to fix it. Any ideas?


One solution is not to call `makeSUT` during tearDown. In the `clearCache` tearDown helper, create a new instance of `CoreDataFeedStore` instead:

```
private func clearCache() {
  deleteCache(from: try! CoreDataFeedStore())
}
```

Another way is to make the `sut` an instance property of the test case and control its lifecycle in setUp/tearDown. For example:

```
class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
    private var sut: CoreDataFeedStore!

    override func setUpWithError() throws {
        super.setUpWithError()
        
        sut = try CoreDataFeedStore()
        deleteCache(from: sut)
    }
    
    override func tearDown() {
        super.tearDown()
        
        deleteCache(from: sut)

        weak var weakSUT: CoreDataFeedStore? = sut
        sut = nil
        XCTAssertNil(weakSUT, "Instance should have been deallocated - potential memory leak")
    }

    func test_retrieve_deliversEmptyOnEmptyCache() {        
        assertThatRetrieveDeliversEmptyOnEmptyCache(on: sut)
    }

    ....
}
```

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,127 @@
...
+//  Created by Nicolas De Maio on 9/20/20.
+//  Copyright © 2020 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+public class CoreDataFeedStore: FeedStore {
+    private let container: NSPersistentContainer?
+    private let context: NSManagedObjectContext?
```
caiozullo, line 14: This class cannot work without an `NSPersistentContainer` or an `NSManagedObjectContext`, so those dependencies shouldn't be optional. If it fails to create them, you could throw an error in the initializer to prevent the class from being created in an invalid state.

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,127 @@
...
+                do {
+                    if let cache: Cache = try context.fetch(Cache.fetchRequest()).first {
+                        var feedImages = [NSManagedObject]()
+                        for localFeedImage in feed {
+                            let feedImage = NSManagedObject(entity: feedImageEntity, insertInto: context)
+                            
+                            feedImage.setValue(localFeedImage.id, forKey: "id")
+                            feedImage.setValue(localFeedImage.description, forKey: "imageDescription")
+                            feedImage.setValue(localFeedImage.location, forKey: "location")
+                            feedImage.setValue(localFeedImage.url, forKey: "url")
```
caiozullo, line 70: You can use the `FeedImage` class here, which is a subclass of `NSManagedObject`. So you can simplify the code using the setters:

```
let feedImage = FeedImage(context: context)
feedImage.id = localFeedImage.id
feedImage.imageDescription = localFeedImage.description
...
```

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,127 @@
...
+                            
+                            feedImages.append(feedImage)
+                        }
+                        
+                        cache.timestamp = timestamp
+                        cache.setValue(NSOrderedSet(array: feedImages), forKey: "items")
```
caiozullo, line 76: You can use the setter instead:

```
cache.items = NSOrderedSet(array: feedImages)
```

FeedStoreChallenge/CoreData Feed Store/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,127 @@
...
+                            feedImage.setValue(localFeedImage.url, forKey: "url")
+                            
+                            feedImages.append(feedImage)
+                        }
+                        
+                        cache.timestamp = timestamp
+                        cache.addToItems(NSOrderedSet(array: feedImages))
+                        
+                        try context.save()
+                        completion(.none)
```
caiozullo, line 99: This code is duplicated - you can unify it

Tests/FeedStoreChallengeTests.swift
```diff
@@ -7,150 +7,108 @@ import FeedStoreChallenge
...
+		assertThatSideEffectsRunSerially(on: sut)
 	}
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT(file: StaticString = #file, line: UInt = #line) -> FeedStore {
+        let sut = try! CoreDataFeedStore()
+        //trackForMemoryLeak(sut, file: file, line: line)
```
caiozullo, line 98: Uncomment this line

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -31,60 +31,67 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+
+        expect(storeToLoad, toRetrieve: .empty)
     }
     
     // - MARK: Helpers
     
     private func makeSUT() -> FeedStore {
-        fatalError("Must be implemented")
+        let sut = try! CoreDataFeedStore()
+        return sut
```
caiozullo, line 80: Need to track for memory leaks

caiozullo (commented):
Hi @Nico113,

- You added the diagram.
- Bonus points for the Integration Tests.

I left a few suggestions and comments that can help you improve the code. Feel free to push changes to your implementation.

Great job ✅

Nico113:
Thanks @caiozullo for your review.
I improved the code addressing your comments.

caiozullo:
Nice one!

PR #33, opened 2020-09-24
SQLite.swift implementation - Danil

dnlbl:
Hi guys! 
I faced with unexpected behavior in comparing two timestamps in the tests, so I changed Date values to corresponding 'timeIntervalSince1970' values and test are passing. Maybe the problem is in SQLite float representation? Is it okay to change the tests because of framework features?


caiozullo:
Hi @dnlbl,

Your `SQLite` implementation looks great. Good job!

To pass the challenge, you shouldn't alter the test assertions. The `SQLite` implementation needs to return the correct date value matching the test setup.

dnlbl:
> Hi @dnlbl,
> 
> Your `SQLite` implementation looks great. Good job!
> 
> To pass the challenge, you shouldn't alter the test assertions. The `SQLite` implementation needs to return the correct date value matching the test setup.

dnlbl:
@caiozullo Thanks for reply!
I found one solution to pass tests without altering test assertions. But maybe there's a better way how to solve it. 
Waiting for your comments!

FeedStoreChallenge/SQLiteFeedStore/SQLiteFeedStore.swift
```diff
@@ -0,0 +1,119 @@
...
+        self.connection = connection
+        self.feedTable = Table("feed")
+        self.workingQueue = DispatchQueue(
+            label: "SQLiteFeedStore",
+            qos: .utility, attributes: .concurrent)
+        prepareTables()
+    }
+    
+    private func prepareTables() {
+        _ = try? connection.run(feedTable.create(block: SQLiteFeedImageHelper.prepareTable))
```
caiozullo, line 37: It'd be best to handle errors here or throw errors so clients can handle them. Otherwise, this instance may be created in an invalid state.

caiozullo (commented):
Hi @dnlbl,

- You added the diagram.
- Bonus points for the Integration Tests.

I left only one suggestion regarding error handling. Feel free to push changes to your implementation.

Great job ✅

FeedStoreChallenge/SQLiteFeedStore/SQLiteFeedStore.swift
dnlbl: Hi @caiozullo! I've added some changes according to your recommendation, thanks for the feedback!

caiozullo: Nice one!

PR #34, opened 2020-10-01
CoreData Implementation - Yannic Borgfeld

Cronay:
For the sake of trying it out I implemented the failable specs in a separate [branch](https://github.com/Cronay/ios-lead-essentials-feed-store-challenge/tree/failable_specs) but then decided it's not worth merging into the final solution because the test details leaking into production code were to big I think.


FeedStoreChallenge/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,115 @@
...
+import Foundation
+import CoreData
+
+public final class CoreDataFeedStore: FeedStore {
+
+    private let persistentContainer: NSPersistentContainer
+    private let context: NSManagedObjectContext
+
+    public init(storeURL: URL) throws {
+        let managedObjectModel = try NSManagedObjectModel.loadFeedStoreModel()
```
caiozullo, line 18: Every time the `CoreDataFeedStore` is instantiated, it creates a new instance of the `NSManagedObjectModel`. But loading a `NSManagedObjectModel` more than once generates console warnings such as:

```
[error] warning: Multiple NSEntityDescriptions claim the NSManagedObject subclass 'ManagedCache' so +entity is unable to disambiguate.
```

You can solve it by making a static `NSManagedObjectModel` instance, so it's loaded *only once* lazily.

caiozullo (commented):
Hi @Cronay,

- Bonus points for the Integration Tests, Memory Leak Tracking, and Failable Specs.

I left only one comment regarding a console warning. Feel free to push a fix.

Great job ✅

FeedStoreChallenge/CoreDataFeedStore/CoreDataFeedStore.swift
Cronay: Thanks for the heads up, I fixed it 👍

caiozullo:
Nice one!

PR #35, opened 2020-10-03
"Realm implementation - Paul Lee"

ctwdtw:
Hi Caio & Mike:
All the tests are passing on my mac with Xcode 12.0.1, macOS 10.15.7.
Look forward to your feedback, thank you : )


FeedStoreChallenge/RealmFeedStore+FeedStore.swift
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+//MARK: - helpers
+private extension Array where Element == RealmFeedImage {
+    func toLocals() -> [LocalFeedImage] {
+        let localFeed = map { LocalFeedImage(id: UUID(uuidString: $0.id)!,
+                                             description: $0.imageDescription,
+                                             location: $0.location,
+                                             url: URL(string: $0.url)!) }
```
caiozullo, line 76: It'd be best to handle the optional cases where the UUID and URL can't be created - otherwise, the app would crash.

caiozullo (commented):
Hi @ctwdtw,

- Bonus points for the Integration Tests, Memory Leak Tracking, and Failable Specs.

I left only one suggestion to prevent a potential crash. Feel free to push a fix.

Great job ✅

caiozullo:
Hi Paul, nice one! 

Replying to your message:

> Hi Caio & Mike:
Thank you for the feedback. For exercise cause, I pushed a [new commit](https://github.com/essentialdevelopercom/ios-lead-essentials-feed-store-challenge/pull/35/commits/83f6a7bcab94284be71f76bbcb09970778ea90ac) that follows the TDD process by writing a test that causes a crash, and then fix it.
However, I doubt if that force-unwraps ( `UUID(uuidString: $0.id)!`, `URL(string: $0.url)!` ) was a problem which needs to be addressed:
The `RealmFeedImage` instance was transferred from a `LocalFeedImage` instances which have a definite-valid `UUID` and a definite-valid `URL` (thanks for the compiler). Therefore when we transfer instances of `RealmFeedImage` back to instances of `LocalFeedImage` the `id`, and `url` properties of a `RealmFeedImage` instance can only be a valid parameter for `UUID` and `URL` initializer respectively, and will not cause any nil value. Since it will never be a nil value, force unwrapped in this scenario is safe, and it is not necessary to "complicate"  codes with `guard let` clauses. Am I right or am I missing something? 
A follow-up question is in what kind of scenario using force unwrap in prod. code is a safe practice? I heard somewhere in the courses you said something like `this would be a developer error, therefore we can force unwrap it`, can you illustrate this a little bit more in a systematic way like what are the different kinds of error or provide some references, thank you. Paul Lee

The problem is that many scenarios could lead to invalid data in the `RealmFeedImage`.

For example, the `RealmFeedImage` defines initializers that take `String` values in both `UUID` and `URL` params:

```
convenience init(id: String, imageDescription: String?, location: String?, url: String) {
``` 

So any client could easily create and store invalid instances.

Also, the default initializer with no params - `RealmFeedImage()` - would lead to invalid instances since the default values for the UUID and URL is empty string:

```
@objc dynamic private(set) var id: String = ""
@objc dynamic private(set) var url: String = ""
```

Other scenarios can also lead to invalid state and runtime crashes, such as schema migrations. 

So since the data is dynamic as has many points of failure, this should be handled in code for this challenge.

If you want to avoid the guard checks, you need to define better APIs in the `RealmFeedImage` to prevent mistakes. For example, you could (1) give better default values, (2) hide the fact that `Realm` stores the values as strings by making the `String` properties private, and (3) pass UUID and URL in the initializer:

```
class RealmFeedImage: Object {
  @objc dynamic private var _id: String = UUID().uuidString
  @objc dynamic private var _url: String = ...
  ...

  var id: UUID {
    UUID(uuidString: _id)!
  }

  var url: URL {
    URL(string: _url)!
  }

  convenience init(id: UUID, url: URL, ...) {
    self.init()
    self._id = id.uuidString
    self._url = url.absoluteString
    ...
  }
}
```

Now, the compiler can prevent creating invalid `RealmFeedImage` instances. Although migrations can still lead to crashes at runtime.

So, it can be ok to force-unwrap when you're **100% sure** it will never fail in prod. For example, when creating a static URL:

```
URL(string: "http://a-valid-url.com")!
```

However, if the URL is created from dynamic string values, it should be handled:

```
guard let url = URL(string: aDynamicString) ...
```

You can also force-unwrap in cases where you're **100% sure** it'd be a developer mistake, and that this mistake would be caught by tests - so it'd never happen in prod. And even though you can guarantee it, it's good practice to handle Optionals and Errors accordingly (including tests!) because they often lead to future crashes when the code changes.

PR #38, opened 2020-10-20
SQLite implementation - Fabio Cuomo

mulp:
I used sqlite to persist data to disk. I have created a sqlite wrapper and used it to execute queries.

### DIFF
--- a/FeedStoreChallenge/SQLiteDatabaseWrapper.swift
+++ b/FeedStoreChallenge/SQLiteDatabaseWrapper.swift
@@ -58,7 +58,7 @@ public class SQLiteDatabaseWrapper {
         sqlite3_finalize(statement)
     }
     
-    public func cacheEntries() throws -> ([LocalFeedImage], Date?) {
+    public func cacheEntries() throws -> ([LocalFeedImage], Date)? {
         var items = [LocalFeedImage]()
         let queryStatement = "SELECT * FROM FeedImageCache"
         let statement = try prepareStatement(sql: queryStatement)
@@ -80,7 +80,9 @@ public class SQLiteDatabaseWrapper {
             }
         }
         sqlite3_finalize(statement)
-        return (items, timestamp)
+        guard !items.isEmpty,
+              let cacheTimestamp = timestamp else { return nil }
+        return (items, cacheTimestamp)
     }
     
     public func addNewEntry(_ feeds: [LocalFeedImage], timestamp: Date) throws {
--- a/FeedStoreChallenge/SQLiteDatabaseWrapper.swift
+++ b/FeedStoreChallenge/SQLiteDatabaseWrapper.swift
@@ -50,13 +50,15 @@ public class SQLiteDatabaseWrapper {
                                 """
         var statement: OpaquePointer?
         if sqlite3_exec(dbPointer, createTableSQL, nil, &statement, nil) != SQLITE_OK {
-            let errmsg = String(cString: sqlite3_errmsg(dbPointer)!)
-            print("error creating table: \(errmsg)")
+            if let sqlError = sqlite3_errmsg(dbPointer) {
+                let errmsg = String(cString: sqlError)
+                throw SQLiteError.prepare(message: errmsg)
+            }
         }
         sqlite3_finalize(statement)
     }
     
-    public func cacheEntries() throws -> ([LocalFeedImage], Date) {
+    public func cacheEntries() throws -> ([LocalFeedImage], Date?) {
         var items = [LocalFeedImage]()
         let queryStatement = "SELECT * FROM FeedImageCache"
         let statement = try prepareStatement(sql: queryStatement)
@@ -78,7 +80,7 @@ public class SQLiteDatabaseWrapper {
             }
         }
         sqlite3_finalize(statement)
-        return (items, timestamp ?? Date())
+        return (items, timestamp)
     }
     
     public func addNewEntry(_ feeds: [LocalFeedImage], timestamp: Date) throws {
--- a/FeedStoreChallenge/SQLiteDatabaseWrapper.swift
+++ b/FeedStoreChallenge/SQLiteDatabaseWrapper.swift
@@ -1,0 +1,141 @@
+//
+//  SQLiteDatabaseWrapper.swift
+//  FeedStoreChallenge
+//
+//  Created by Fabio Cuomo on 19/10/2020.
+//  Copyright © 2020 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import SQLite3
+
+public enum SQLiteError: Error {
+    case openDatabase(message: String)
+    case createTable(message: String)
+    case prepare(message: String)
+    case query(message: String)
+    case unknownError
+}
+
+public class SQLiteDatabaseWrapper {
+    private var dbPointer: OpaquePointer? = nil
+    internal let SQLITE_TRANSIENT = unsafeBitCast(-1, to: sqlite3_destructor_type.self)
+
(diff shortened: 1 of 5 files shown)


Tests/FeedStoreChallengeTests.swift
```diff
@@ -7,6 +7,19 @@ import FeedStoreChallenge
 
 class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
 	
+    private let fileURL = try! FileManager.default
+        .url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
+        .appendingPathComponent("feedStoreCache.sqlite")
```
caiozullo, line 12: Access to the `. documentDirectory` requires user permission - for test purposes, use the cache directory to avoid permission issues in the CI server.

caiozullo (commented):
Hi @mulp,

The CI build is failing - I believe it's because the tests are trying to store data to the documents directory but they don't have access to it. Try using the caches directory instead.

Let me know when you solve it. Thank you ✅

Tests/FeedStoreChallengeTests.swift
mulp: ops, forgot it. I'm going to change and update. Thanks

FeedStoreChallenge/SQLiteDatabaseWrapper.swift
```diff
@@ -0,0 +1,152 @@
...
+                                  description VARCHAR(255),
+                                  location VARCHAR(255),
+                                  url VARCHAR(255),
+                                  timestamp REAL
+                                );
+                                """
+        var statement: OpaquePointer?
+        if sqlite3_exec(dbPointer, createTableSQL, nil, &statement, nil) != SQLITE_OK {
+            let errmsg = String(cString: sqlite3_errmsg(dbPointer)!)
+            print("error creating table: \(errmsg)")
```
caiozullo, line 54: It'd be best to handle or throw errors here to avoid issues at runtime. Also, avoid the `sqlite3_errmsg` force unwrap.

FeedStoreChallenge/SQLiteDatabaseWrapper.swift
```diff
@@ -0,0 +1,152 @@
...
+                                      description: string(from: description),
+                                      location: string(from: location),
+                                      url: url)
+            items.append(item)
+            if timestamp == nil {
+                timestamp = Date(timeIntervalSinceReferenceDate: sqlite3_column_double(statement, 4))
+            }
+        }
+        sqlite3_finalize(statement)
+        return (items, timestamp ?? Date())
```
caiozullo, line 81: The `timestamp` should be created along with the other properties in the guard statement so if the timestamp is `nil`, it throws an error. 

The store implementation shouldn't create a default timestamp (Date()) because the timestamp is decided by a higher-level module. So the store shouldn't make this decision.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -7,150 +7,99 @@ import FeedStoreChallenge
...
-	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
-	}
-	
+	private func makeSUT() -> FeedStore? {
+        do {
+            return try SQLiteFeedStore(dbURL: fileURL)
+        } catch {
+            return nil
+        }
```
caiozullo, line 103: You could use `try?` instead of the do/catch. You can also use `try!` and return a non-optional `FeedStore`  to avoid the force-unwrap in every test method.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -31,60 +35,64 @@ class FeedStoreIntegrationTests: XCTestCase {
...
     // - MARK: Helpers
     
-    private func makeSUT() -> FeedStore {
-        fatalError("Must be implemented")
+    private func makeSUT() -> FeedStore? {
+        do {
+            return try SQLiteFeedStore(dbURL: fileURL)
+        } catch {
+            return nil
+        }
```
caiozullo, line 87: You could use `try?` instead of the do/catch. You can also use `try!` and return a non-optional `FeedStore`  to avoid the force-unwrap in every test method.

caiozullo (commented):
Hi @mulp,

All tests are passing, so your implementation complies with the specs.

I left a few comments that can help you improve the implementation. Feel free to push changes.

Great job ✅

FeedStoreChallenge/SQLiteDatabaseWrapper.swift
mulp: Hey @caiozullo thanks for your suggestion. I agree that the store shouldn't create a default timestamp, so because when the cache is empty the timestamp value doesn't matter, I left it as optional. 
Later in the higher-lever module, when there are some entries in the cache I check for the existence and in the case it's nil I return a different domain error.

What do you think?

caiozullo: Hi @mulp, the timestamp would only be `nil` if there's no cache in the DB. In this case, the method should return `nil` and the operation should complete with `.empty` instead of throwing errors.

So the result of the `cacheEntries` method should be optional (instead of an empty array and an optional date):

```
public func cacheEntries() throws -> ([LocalFeedImage], Date)?
```

When there's no cache in the DB, you can return `nil`. Thus, you don't need to check if the entries array is empty anymore:

```
let (entries, timestamp) = try db.cacheEntries()
if entries.isEmpty { ❌
  completion(.empty)
}
```

Instead, you check if there is a value in the cache:

```
if let (entries, timestamp) = try db.cacheEntries() { ✅
  completion(.found(feed: entries, timestamp: timestamp))
} else {
  completion(.empty)
} 
```

This way, you avoid throwing errors for an invalid state in the system.

mulp: Better, I like it. I applied your suggestions. Just one thing: because 
- when I get the value for the timestamp I use sqlite3_column_double which doesn't return an optional. So I cannot include it in the guard statement. I get it only once and before to return I use another guard statement to check what I have to return. Is not the best elegant way...what do you think?

caiozullo: Nice one! You can even remove the `!items.isEmpty` check.

FeedStoreChallenge/SQLiteFeedStore.swift
```diff
@@ -0,0 +1,55 @@
...
+        do {
+            try db.clearCache()
+            completion(nil)
+        } catch let error {
+            completion(error)
+        }
+    }
+    
+    public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+        deleteCachedFeed { [db] error in
```
caiozullo, line 33: Looks like the deletion error is not handled here - it might be best to use `try db.clearCache()` inside the `do` block, so deletion errors and handled within the `catch` block.

FeedStoreChallenge/SQLiteDatabaseWrapper.swift
```diff
@@ -0,0 +1,156 @@
...
+    private var errorMessage: String {
+        if let errorPointer = sqlite3_errmsg(dbPointer) {
+            let errorMessage = String(cString: errorPointer)
+            return errorMessage
+        } else {
+            return "No error message provided from sqlite."
+        }
+    }
+    
+    private init(dbPointer: OpaquePointer?) {
```
caiozullo, line 33: The `dbPointer` shouldn't be optional because this class wouldn't work without one.

PR #42, opened 2020-10-30
InMemory implementation - Armand van der Zwan

armandzwan:


FeedStoreChallenge/InMemoryFeedStore.swift
```diff
@@ -0,0 +1,45 @@
...
+    private struct Cache {
+        let feedItems: [LocalFeedImage]
+        let timestamp: Date
+    }
+    
+    public init() {
+        
+    }
+    
+    private var localStore: [Cache] = []
```
caiozullo, line 21: Since we only store a single cache value, it could be an `Optional` instead of an `Array`:

```suggestion
    private var cache: Cache?
```

caiozullo (commented):
Hi @armandzwan,

All tests are passing, so your implementation complies with the specs.

I left one suggestion about using `Optional` instead of `Array`. Feel free to push changes.

Great job! ✅

armandzwan:
Yes, good point. It was because I started thinking of creating a Realm or Core Data implementation. Optional is good enough for an in-memory solution.

caiozullo:
Nice one!

PR #44, opened 2020-11-11
Realm Implementaion - Ángel Vázquez

Angel5215:


Tests/FeedStoreChallengeTests.swift
```diff
@@ -18,139 +18,112 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+    
+    override func setUp() {
+        super.setUp()
+        try? FileManager.default.removeItem(at: testStoreURL())
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+        try? FileManager.default.removeItem(at: testStoreURL())
+    }
```
caiozullo, line 30: You could use an in-memory Realm configuration to speed up these tests, as it wouldn't write to disk. You can use a normal configuration only for the integration tests.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -18,139 +18,112 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
-//		let sut = makeSUT()
-//
-//		assertThatSideEffectsRunSerially(on: sut)
+		let sut = makeSUT()
+
+		assertThatSideEffectsRunSerially(on: sut)
 	}
+    
+    func test_init_throwsErrorWithInvalidConfiguration() throws {
+        let configuration = makeConfiguration(url: cachesURL())
```
caiozullo, line 105: It's not clear why the `cachesURL` is invalid. You can give it a better name such as `invalidURL()` or a new method `invalidConfiguration()` to clarify.

caiozullo (commented):
Hi @Angel5215,

- All tests are passing ✅ 

I left minor suggestions. Feel free to push changes.

Great job!

caiozullo:
Nice one!

PR #52, opened 2020-11-27
timestamp assertion with accuracy

ytyubox:
I notice that Compare between Dates somehow requires more accuracy than actually need.
Some Persistence API reduces accuracy. I found this on trying `GRDB.swift`.

```swift
precondition(retrievedTimestamp == expectedTimestamp, "retreved: \(retrievedTimestamp.timeIntervalSince1970), expect: \(expectedTimestamp.timeIntervalSince1970)")
```
fail with this message:
> Precondition failed: retreved: 1606465034.644723, expect: 1606465034.644723


caiozullo:
Hi @ytyubox, thank you for the PR!

Storing the values without losing accuracy is part of the challenge. So the current assertion is necessary.

groue:
@ytyubox, you can get lossless Date roundtrip with GRDB. When you deal with raw database rows, store doubles and decode doubles (with `date.timeIntervalSinceReferenceDate` and `Date.init(timeIntervalSinceReferenceDate:)`). When you use Codable record types, use the `. timeIntervalSinceReferenceDate` date encoding/decoding strategies.

By default, GRDB stores dates in a format that favors SQLite use cases (database queries, SQL date functions, data exchange with other platforms, etc.) Exact value roundtrip, unfortunately, hinders those use cases. But you can escape the default behavior when you need, as explained above. See the [Date and DateComponents](https://github.com/groue/GRDB.swift/blob/master/README.md#date-and-datecomponents) chapter for more information).

ytyubox:
@groue Thank you so much. You save my day.
Just wondering, how did you find this pr? Because literally I just only use the plain string in the description. Really appreciated.

groue:
@ytyubox, I sometimes search GitHub for GRDB-related issues, just for inspiration: https://github.com/search?o=desc&q=GRDB&s=updated&type=Issues

PR #55, opened 2020-11-30
Realm implementation - Jair Moreno

JAIRMG:
Hi Mike and Caio, I've been enjoying so much this course 👍. I would like to share to you something that I found doing this challenge:
1. During the implementation I chose SPM to use Realm because this was going to be my first time using it, however, I encounter multiple warnings regarding **symlinks duplicated** I tried to remove those warnings but I couldn't find a solution. So I change the implementation to Cocoapods.
2. At the beginning I implemented the concurrent call as we did in Essential Feed, however I realized that the behavior of Realm during this insertion, retrieval and deletion was working fine even without those barriers, so I removed them. This also helps me to inject a configuration of realm and avoid creating a Realm instance in every method. This way, I got rid of the **Realm Objects cannot be accessed from different threads** error too. 
3. I could not find a way to test the _failable tests_ for a reason: everytime I inject a wrong configuration an error was thrown preventing me to reach the **write** step, so I made a test where I validated that giving the wrong config an error was thrown while creating the SUT. I hope this is ok.

Thanks for making such a great tutorials and taking the time to response to everyone doubts, you guys rock 🤘


Tests/FeedStoreChallengeTests.swift
```diff
@@ -4,153 +4,125 @@
...
+        configuration.fileURL = invalidURL()
+
+        XCTAssertThrowsError(try RealmFeedStore(configuration: configuration))
+    }
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
+    private func makeSUT(configuration: Realm.Configuration = Realm.Configuration()) -> FeedStore {
```
caiozullo, line 102: You can run those tests with an in-memory configuration to speed up the tests. You also won't need to undo side-effects.

You can use the default configuration with side-effects in the integration tests only.

FeedStoreChallenge/RealmFeedStore/RealmFeedImage.swift
```diff
@@ -0,0 +1,36 @@
...
+        self.cacheDescription = image.description
+        self.location = image.location
+        self.url = image.url.absoluteString
+    }
+    
+    var local: LocalFeedImage {
+        .init(id: UUID(uuidString: id)!,
+              description: cacheDescription,
+              location: location,
+              url: URL(string: url)!)
```
caiozullo, line 33: Ideally, the UUID/URL should be created without the `!` to prevent runtime errors. You could return `nil` and discard this record if the data cannot be converted to appropriate values.

caiozullo (commented):
Hi @JAIRMG,

- All tests are passing ✅ 

I left minor review suggestions. Feel free to push changes.

Great job!

Regarding thread safety, the `Realm` instance is not thread-safe. It can only be used on the same thread it was created. So this implementation is not thread-safe. It's ok for this challenge, but beware when using it in production! Every thread should have its own `Realm` instance, or you can protect access to it with a queue, for example.

FeedStoreChallenge/RealmFeedStore/RealmFeedStore.swift
```diff
@@ -0,0 +1,85 @@
...
+import RealmSwift
+
+public class RealmFeedStore: FeedStore {
+    
+    private static let cacheId = "cache"
+	private let configuration: Realm.Configuration
+	
+	private let queue = DispatchQueue(label: "\(RealmFeedStore.self)Queue", qos: .userInitiated, attributes: .concurrent)
+    
+    public init(configuration: Realm.Configuration) throws {
```
caiozullo, line 18: The `init` doesn't throw errors anymore - you can remove `throws`

FeedStoreChallenge/RealmFeedStore/RealmFeedStore.swift
```diff
@@ -0,0 +1,85 @@
...
+		self.configuration = configuration
+    }
+    
+    public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		queue.async(flags: .barrier) {
+			
+			do {
+				let realm = try Realm()
```
caiozullo, line 26: When using `Realm` instances in a background queue, it's good practice to wrap it into an explicit autoreleasepool. Learn more at: https://realm.io/docs/swift/latest/#threading

FeedStoreChallenge/RealmFeedStore/RealmFeedStore.swift
```diff
@@ -0,0 +1,85 @@
...
+		}
+    }
+    
+    public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+        
+		queue.async(flags: .barrier) {
+			let feedList = feed.map(RealmFeedImage.init)
+			
+			do {
+				let realm = try Realm()
```
caiozullo, line 49: When using `Realm` instances in a background queue, it's good practice to wrap it into an explicit autoreleasepool. Learn more at: https://realm.io/docs/swift/latest/#threading

FeedStoreChallenge/RealmFeedStore/RealmFeedStore.swift
```diff
@@ -0,0 +1,85 @@
...
+			
+		}
+
+    }
+
+    public func retrieve(completion: @escaping RetrievalCompletion) {
+		queue.async {
+			
+			do {
+				let realm = try Realm()
```
caiozullo, line 71: When using `Realm` instances in a background queue, it's good practice to wrap it into an explicit autoreleasepool. Learn more at: https://realm.io/docs/swift/latest/#threading

Tests/FeedStoreChallengeTests.swift
```diff
@@ -4,153 +4,124 @@
...
+    
+    override func setUp() {
+        super.setUp()
+        setupEmptyState()
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+        undoStoreFeedSideEffects()
+    }
```
caiozullo, line 19: You don't need to undo side-effects in an in-memory database. Remove the setUp/tearDown and the helper methods.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -4,83 +4,92 @@
...
+
+        deleteCache(from: storeToDelete)
+
+        expect(storeToLoad, toRetrieve: .empty)
+    }
+    
+    // - MARK: Helpers
+    
+    private func makeSUT() -> FeedStore {
+        let sut = try! RealmFeedStore(configuration: inMemoryConfigurationRealm())
```
caiozullo, line 69: It'd be best to run the Integration Tests with the default configuration to replicate the production behavior. Use the in-memory one for the unit tests only.

FeedStoreChallenge/RealmFeedStore/RealmFeedStore.swift
JAIRMG: After I removed the setUp/tearDown the first two tests in `RealmFeedStore` were failing, I attribute that issue to the background queue so I tried to wrap it into an explicit autoreleasepool, but I got stuck due to the following: 

this code does not present any error
                 
```
		queue.async(flags: .barrier) { [weak self] in
			guard let self = self else { return }
			autoreleasepool {
				
				let feedList = feed.map(RealmFeedImage.init)
				
				do {
					let realm = try Realm(configuration: self.configuration)
					try realm.write {
						realm.deleteAll()
						let cache = Cache()
						cache.feed.append(objectsIn: feedList)
						cache.timestamp = timestamp
						cache.id = RealmFeedStore.cacheId
						realm.add(cache)
					}
					completion(nil)
				} catch {
					completion(error)
				}
			}
		}
    }

```
however, this code does present an error of type `Unable to infer complex closure return type; add explicit type to disambiguate releasepool`

````
		queue.async { [weak self] in
			guard let self = self else { return }
			autoreleasepool {
				
				do {
					let realm = try Realm(configuration: self.configuration)
					guard let cache = realm.objects(Cache.self).first else {
						return completion(.empty)
					}
					
					completion(.found(feed: cache.localFeed, timestamp: cache.timestamp))
				} catch {
					completion(.failure(error))
				}
			}
		}
    }
````
My question is, how can I accomplish that? Is this what you meant? I send my changes removing the queue, this way is working fine but I am intrigued

caiozullo: Hi @JAIRMG, the error message is describing the problem. The compiler can't infer the return type. You can set it manually or extract the logic into functions to avoid the nested closures.

JAIRMG: Hi Caio, I extract the logic into separate function and make this work but `autoreleasepool` inside background thread is causing memory leak

caiozullo: Memory leak where? `autoreleasepool` shouldn't generate memory leaks here.

Please, follow the Realm docs if you're having issues.

JAIRMG: I also don't understand why should we wrap the logic into an `autoreleasepool` if we are not sharing any instance of Realm, from doc:
`The only thing you have to be aware of is that you cannot have multiple threads sharing the same instances of Realm objects.` I would appreciate if you can help me with this doubt

caiozullo: Hi @JAIRMG, you don't need to make the store thread-safe for this challenge. You already passed.

But if you're going to make it thread-safe, it's important to follow the Realm docs. For example, from the docs: *Avoid “pinning” old Realm transactions by using auto-refreshing Realms and wrapping all use of Realm APIs from background threads in explicit autorelease pools.*

Without an explicit autorelease pool, the Realm instance may outlive the operation scope.

You can see in the docs they wrap every operation in background queues with explicit autorelease pools.

JAIRMG: Thank you for taking the time to respond and review multiple times the PR. I really I appreciate 👍 Definitely will dive deep into this topic

caiozullo: Keep up the great work!

Learn, practice, and execute 🙂

PR #56, opened 2020-12-02
Realm implementation - Adrian Szymanowski

Adrimi:
One note about five compile-time warnings:
I am aware of that I violated guideline rule no5, but from I've found out Realm database community confirmed that they are working on that problem (and full support on Swift Package Manager). https://github.com/realm/realm-cocoa/issues/6923


caiozullo:
Hi @Adrimi, to pass the challenge there shouldn't be any warnings. I believe other students used Cocoapods to integrate Realm because of the warning issues they have at the moment with SPM.

PR #57, opened 2020-12-03
GRDB.swift  implementation - Tsungyu Yu

ytyubox:
Hi Caio and Mike,

This is my pull request for Feed Store Challenge, my second attempt in GRDB.swift.
Please have a look, thanks. 😁


FeedStoreChallenge/GRDBFeedStore/GRDBFeedDataBase.swift
```diff
@@ -0,0 +1,71 @@
...
+//
+//  Created by 游宗諭 on 2020/12/3.
+//  Copyright © 2020 Essential Developer. All rights reserved.
+//
+
+import GRDB
+public protocol DBQueueManager {
+    func makeQueue(path: String?) throws -> DatabaseQueue
+    func recylcle(queue: DatabaseQueue)
+}
```
caiozullo, line 13: This abstraction seems unnecessary. You could remove it.

FeedStoreChallenge/GRDBFeedStore/GRDBFeedDataBase.swift
```diff
@@ -0,0 +1,71 @@
...
+
+public class DBQueueFactory: DBQueueManager {
+    public init() {}
+    private var pool:[String?: (count:Int,dbQueue: DatabaseQueue)] = [:]
+    public func makeQueue(path: String?) throws -> DatabaseQueue {
```
caiozullo, line 18: Can the `path` be nil? How would it work?

Moreover, the `path` is being used to create the `DatabaseQueue`, so it may not work as intended.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -6,81 +6,111 @@ import XCTest
...
+        let storeToInsert = makeSUT(dbQueueManager: manager)
+        let storeToDelete = makeSUT(dbQueueManager: manager)
+        let storeToLoad = makeSUT(dbQueueManager: manager)
+
+        insert((uniqueImageFeed(), Date()), to: storeToInsert)
+
+        deleteCache(from: storeToDelete)
+
+        expect(storeToLoad, toRetrieve: .empty)
+    }
```
caiozullo, line 72: To pass the challenge, the tests shouldn't be modified. Revert the change in the tests and implement only the `makeSUT` function.

FeedStoreChallenge/GRDBFeedStore/GRDBFeedStore.swift
```diff
@@ -0,0 +1,85 @@
...
+    
+    public init(path: String?, dbManeger: DBQueueManager) throws {
+        self.dbManeger = dbManeger
+        dbQueue = try dbManeger.makeQueue(path: path)
+    }
+    deinit {
+        dbManeger.recylcle(queue: dbQueue)
+    }
+    public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+        try? dbQueue.write({ (db) in
```
caiozullo, line 27: Errors should be handled. But `try?` will ignore errors.

FeedStoreChallenge/GRDBFeedStore/GRDBFeedDataBase.swift
ytyubox: In the library, if we want to use a in memory database, we can use the hard code path: `:memory:`. 
And yes, I did not use the path for my DatabaseQueue, and its fix by next commit, thank you.

FeedStoreChallenge/GRDBFeedStore/GRDBFeedDataBase.swift
ytyubox: Yes, totally agree.

Tests/FeedStoreIntegrationTests.swift
ytyubox: The DBQueueFactory now become a Static member (let) of GRDBFeedStore.

FeedStoreChallenge/GRDBFeedStore/GRDBFeedStore.swift
ytyubox: Sure, the issue has been fixed by follow up commit.

caiozullo (commented):
Hi @ytyubox,

- All tests are passing ✅ 

Great job!

PR #59, opened 2020-12-07
Realm feed store

Adrimi:
This solution does not produce any warnings only on Xcode version up to 11.7. Newer version of Xcode (12.X) does produce several errors caused by Realm framework not officially updated to newest Xcode release.


caiozullo:
Hi @Adrimi, if you're using Xcode 11, you should use the xcode11 branch. Please, check the instructions in the readme file. Thank you!

Adrimi:
Ok, I changed cocoa pods to Dynamic frameworks implementation. Looks like no warnings and tests are passing :D

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -4,83 +4,85 @@
...
+        return realm
+    }
+    
+    private func setupEmptyStoreState() {
+        Realm.Configuration.defaultConfiguration.inMemoryIdentifier = self.name
+    }
+
+    private func undoStoreSideEffects() {
+        Realm.Configuration.defaultConfiguration.inMemoryIdentifier = self.name
+    }
```
caiozullo, line 87: You can use the default config (not in-memory) for the integration tests to check the whole integration.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -4,153 +4,109 @@
...
+    //
+    //  ***********************
+    
+    override func setUp() {
+        super.setUp()
+        // Use an in-memory Realm identified by the name of the current test.
+        // This ensures that each test can't accidentally access or modify the data
+        // from other tests or the application itself, and because they're in-memory,
+        // there's nothing that needs to be cleaned up.
+        Realm.Configuration.defaultConfiguration.inMemoryIdentifier = self.name
```
caiozullo, line 29: It'd be best to set the `inMemoryIdentifier` in a copy of this config instead of setting it globally.

I recommend you remove this `setUp` method and create the configuration in the `makeSUT`.

FeedStoreChallenge/RealmFeedImage.swift
```diff
@@ -0,0 +1,40 @@
...
+
+class RealmFeedImage: Object {
+    @objc dynamic var id: String = ""
+    @objc dynamic var imageDescription: String? = nil
+    @objc dynamic var location: String? = nil
+    @objc dynamic var urlString: String = ""
+    
+    override init() {}
+    
+    init(id: String, imageDescription: String?, location: String?, urlString: String) {
```
caiozullo, line 20: It'd be best to set the right types to prevent mistakes. id: UUID, URL: URL

caiozullo (commented):
Hi @Adrimi,

- All tests are passing ✅ 

Great job!

I left some review suggestions. Feel free to push changes.

Please, *update your PR name following the instructions in the read me file.* Thank you 🙂

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -4,83 +4,81 @@
...
+    // - MARK: Helpers
+    
+    private func makeSUT() -> FeedStore {
+		let configuration = Realm.Configuration.defaultConfiguration
+        let realm = try! RealmFeedStore(configuration: configuration)
+        return realm
+    }
+    
+	private func clearCache() {
+		makeSUT().deleteCachedFeed(completion: {_ in})
```
caiozullo, line 82: It'd be best to wait for the completion block to avoid issues (e.g., other tests running before the operation completes).

You can use the `deleteCache` test helper:

```
deleteCache(from: makeSUT())
```

Adrimi: Oh sure, I would never found that by myself :D Thanks (btw I needed to change method name to `deleteCachedData` since `deleteCache(from:)` is already in the namespace

caiozullo: Hi @Adrimi, you can use the one that exists already. No need to create a new one.

Adrimi: Right, I've missed that. Thanks again

caiozullo:
Nice one!

PR #64, opened 2020-12-31
Realm implementation - Joseph Wardell

jaywardell:
Here's my Realm solution to the feed store challenge.

I ran across a couple interesting issues:
1) Realm doesn't like Realm.Object subclasses that exist within a namespace other than the global namespace.  So I was unable to include my RealmFeedStoreCachedFeedImage class within my RealmFeedStore implementation as I would have liked.  I was not even able to mark it as fileprivate without getting an error from Realm.  Instead, I marked all properties and methods on it private and left a warning in the documentation for the class.  I was unable to think of a more complete approach to prevent client code from accessing my internal subclass.

2) To test failure on insertion, I tried using a custom Realm.Configuration that was set up as read only.  This worked most of the time, but occasionally I would get a runtime crash for an uncaught NSException.  It seems that Realm's beginWriting() method doesn't catch all NSExceptions (at least that's my current guess).  To deal with this, I added an Objective C bridging header and a little helper OjectiveCExceptions class that would catch NSExceptions thrown from Realm.beginWriting().  I'm curious if there's a better way.

### DIFF
--- a/BridgingHeader.h
+++ b/BridgingHeader.h
@@ -1,0 +1,24 @@
+//
+//  BridgingHeader.h
+//  FeedStoreChallenge
+//
+//  Created by Joseph Wardell on 12/31/20.
+//  Copyright © 2020 Essential Developer. All rights reserved.
+//
+
+@import Foundation;
+
+#ifndef BridgingHeader_h
+#define BridgingHeader_h
+NS_INLINE NSException * _Nullable ExecuteWithObjCExceptionHandling(void(NS_NOESCAPE^_Nonnull tryBlock)(void)) {
+	@try {
+		tryBlock();
+	}
+	@catch (NSException *exception) {
+		return exception;
+	}
+	return nil;
+}
+
+
+#endif /* BridgingHeader_h */
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -7,21 +7,69 @@
 //
 
 import Foundation
+import RealmSwift
 
+class CachedFeedImage: Object {
+	@objc dynamic var id: String?
+	@objc dynamic var desc: String?
+	@objc dynamic var location: String?
+	@objc dynamic var url: String?
+	@objc dynamic var timestamp: Date?
+
+	class func createFeedImage(from localFeedImage: LocalFeedImage, at timestamp: Date) -> CachedFeedImage {
+		
+		let out = CachedFeedImage()
+		out.id = localFeedImage.id.uuidString
+		out.desc = localFeedImage.description
+		out.location = localFeedImage.location
+		out.url = localFeedImage.url.absoluteString
+		
+		out.timestamp = timestamp
+		
+		return out
+	}
+		
+	var localFeedImage: LocalFeedImage {
+		LocalFeedImage(id: UUID(uuidString: id!)!, description: desc, location: location, url: URL(string: url!)!)
+	}
+}
+
+// MARK:-
 public final class RealmFeedStore: FeedStore {
 	
-	public init() {}
+	let realm: Realm
+	public init() {
+		self.realm = try! Realm()
+	}
 	
+
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
 		fatalError()
 	}
 	
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
-		fatalError()
+		try! realm.write {
+			for image in feed {
+				realm.add(CachedFeedImage.createFeedImage(from: image, at: timestamp))
+			}
+		}
+		completion(nil)
 	}
 	
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		completion(.empty)
+		let cached = realm.objects(CachedFeedImage.self)
+		if cached.count > 0 {
+			let feedImages = cached.map {
+				$0.localFeedImage
+			}
+			
+			let timestamp = cached.first!.timestamp!
+			
+			completion(.found(feed: Array(feedImages), timestamp: timestamp))
+		}
+		else {
+			completion(.empty)
+		}
 	}
 	
 	
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -54,8 +54,8 @@ public final class RealmFeedStore: FeedStore {
 	
 	let configuration: Realm.Configuration
 	let queue: DispatchQueue
-	public init(_ fileURL: URL) {
-		self.configuration = Realm.Configuration(fileURL: fileURL)
+	public init(_ fileURL: URL, readOnly: Bool = false) {
+		self.configuration = Realm.Configuration(fileURL: fileURL, readOnly: readOnly)
 		self.queue = DispatchQueue(label: "\(type(of: Self.self))", qos: .userInitiated, autoreleaseFrequency: .workItem)
 	}
 	
@@ -71,43 +71,51 @@ public final class RealmFeedStore: FeedStore {
 		}
 	}
 	
-	private func writeToRealm(_ callback: @escaping (Realm?)->()) throws {
+	private func writeToRealm(_ callback: @escaping (Result<Realm, Error>)->()) throws {
 		queue.async { [unowned self] in
 			do {
 				let realm = try Realm(configuration: self.configuration, queue: self.queue)
 				try realm.write {
-					callback(realm)
+					callback(.success(realm))
 				}
 			}
 			catch {
-				
+				callback(.failure(error))
 			}
 		}
 	}
 	
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
 
-		try! writeToRealm { realm in
-			guard let realm = realm else { return completion(nil) }
-
-			realm.deleteAll()
-			completion(nil)
+		try! writeToRealm {
+			switch $0 {
+			case .failure(let error):
+				completion(error)
+			
+			case .success(let realm):
+				realm.deleteAll()
+				completion(nil)
+			}
 		}
 	}
 	
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
 		
-		try! writeToRealm { realm in
-			guard let realm = realm else { return completion(nil) }
-
-			realm.deleteAll()
-
-			for image in feed {
-				realm.add(RealmFeedStoreCachedFeedImage.create(from: image, at: timestamp))
+		try! writeToRealm {
+			switch $0 {
+			case .failure(let error):
+				completion(error)
+				
+			case .success(let realm):
+				realm.deleteAll()
+				
+				for image in feed {
+					realm.add(RealmFeedStoreCachedFeedImage.create(from: image, at: timestamp))
+				}
+				realm.add(RealmFeedStoreTimestamp.create(from: timestamp))
+				
+				completion(nil)
 			}
-			realm.add(RealmFeedStoreTimestamp.create(from: timestamp))
-			
-			completion(nil)
 		}
 	}
 	
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -58,42 +58,46 @@ public final class RealmFeedStore: FeedStore {
 		self.configuration = Realm.Configuration(fileURL: fileURL)
 		self.queue = DispatchQueue(label: "\(type(of: Self.self))", qos: .userInitiated)
 	}
-	
-	private func oneUseRealm() throws -> Realm {
-		// from the documentation for Realm.init():
-		// Realm instances are cached internally, and constructing equivalent Realm objects (for example, by using the same path or identifier) produces limited overhead.
-		try Realm(configuration: configuration)
+		
+	private func accessRealm(_ callback: (Realm)->()) throws {
+		try autoreleasepool {
+			let realm = try Realm(configuration: configuration)
+			callback(realm)
+		}
 	}
 	
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
 
-		let realm = try! oneUseRealm()
-		clearOld(from: realm)
-		completion(nil)
+		try! accessRealm { realm in
+			clearOld(from: realm)
+			completion(nil)
+		}
 	}
 	
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
 		
-		let realm = try! oneUseRealm()
-		clearOld(from: realm)
-		try! realm.write {
-			for image in feed {
-				realm.add(RealmFeedStoreCachedFeedImage.create(from: image, at: timestamp))
+		try! accessRealm { realm in
+			clearOld(from: realm)
+			try! realm.write {
+				for image in feed {
+					realm.add(RealmFeedStoreCachedFeedImage.create(from: image, at: timestamp))
+				}
+				
+				realm.add(RealmFeedStoreTimestamp.create(from: timestamp))
 			}
-			
-			realm.add(RealmFeedStoreTimestamp.create(from: timestamp))
+			completion(nil)
 		}
-		completion(nil)
 	}
 	
 	public func retrieve(completion: @escaping RetrievalCompletion) {
 
-		let realm = try! oneUseRealm()
-		guard let timestamp = realm.objects(RealmFeedStoreTimestamp.self).last?.timestamp else { return completion(.empty) }
-
-		let cached = realm.objects(RealmFeedStoreCachedFeedImage.self)
-		let feedImages = Array(cached.map(\.localFeedImage))
-		completion(.found(feed: feedImages, timestamp: timestamp))
+		try! accessRealm { realm in
+			guard let timestamp = realm.objects(RealmFeedStoreTimestamp.self).last?.timestamp else { return completion(.empty) }
+			
+			let cached = realm.objects(RealmFeedStoreCachedFeedImage.self)
+			let feedImages = Array(cached.map(\.localFeedImage))
+			completion(.found(feed: feedImages, timestamp: timestamp))
+		}
 	}
 	
 	private func clearOld(from realm: Realm) {
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -17,23 +17,35 @@ final class RealmFeedStoreCachedFeedImage: Object {
 	@objc fileprivate dynamic var desc: String?
 	@objc fileprivate dynamic var location: String?
 	@objc fileprivate dynamic var url: String?
-	@objc fileprivate dynamic var timestamp: Date?
 
-	fileprivate class func createFeedImage(from localFeedImage: LocalFeedImage, at timestamp: Date) -> RealmFeedStoreCachedFeedImage {
+	fileprivate class func create(from localFeedImage: LocalFeedImage, at timestamp: Date) -> RealmFeedStoreCachedFeedImage {
 		
 		let out = RealmFeedStoreCachedFeedImage()
 		out.id = localFeedImage.id.uuidString
 		out.desc = localFeedImage.description
 		out.location = localFeedImage.location
 		out.url = localFeedImage.url.absoluteString
-		
-		out.timestamp = timestamp
-		
+				
 		return out
 	}
 		
 	fileprivate var localFeedImage: LocalFeedImage {
 		LocalFeedImage(id: UUID(uuidString: id!)!, description: desc, location: location, url: URL(string: url!)!)
+	}
+}
+
+/// This is an internal type to RealmFeedStore
+/// It should not be used by any code outside this file
+// we would make it fileprivate, but Realm complains with an error
+final class RealmFeedStoreTimestamp: Object {
+	@objc fileprivate dynamic var timestamp: Date?
+	
+	fileprivate class func create(from timestamp: Date) -> RealmFeedStoreTimestamp {
+		
+		let out = RealmFeedStoreTimestamp()
+		out.timestamp = timestamp
+		
+		return out
 	}
 }
 
@@ -53,8 +65,10 @@ public final class RealmFeedStore: FeedStore {
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
 		try! realm.write {
 			for image in feed {
-				realm.add(RealmFeedStoreCachedFeedImage.createFeedImage(from: image, at: timestamp))
+				realm.add(RealmFeedStoreCachedFeedImage.create(from: image, at: timestamp))
 			}
+			
+			realm.add(RealmFeedStoreTimestamp.create(from: timestamp))
 		}
 		completion(nil)
 	}
@@ -66,7 +80,7 @@ public final class RealmFeedStore: FeedStore {
 				$0.localFeedImage
 			}
 			
-			let timestamp = cached.first!.timestamp!
+			let timestamp = realm.objects(RealmFeedStoreTimestamp.self).first!.timestamp!			
 			
 			completion(.found(feed: Array(feedImages), timestamp: timestamp))
 		}
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -59,7 +59,7 @@ public final class RealmFeedStore: FeedStore {
 	
 	
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError()
+		completion(nil)
 	}
 	
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -52,22 +52,30 @@ final class RealmFeedStoreTimestamp: Object {
 // MARK:-
 public final class RealmFeedStore: FeedStore {
 	
-	let realm: Realm
+	let configuration: Realm.Configuration
+	let queue: DispatchQueue
 	public init(_ fileURL: URL) {
-		let configuration = Realm.Configuration(fileURL: fileURL)
-		self.realm = try! Realm(configuration: configuration)
+		self.configuration = Realm.Configuration(fileURL: fileURL)
+		self.queue = DispatchQueue(label: "\(type(of: Self.self))", qos: .userInitiated)
 	}
 	
+	private func oneUseRealm() throws -> Realm {
+		// from the documentation for Realm.init():
+		// Realm instances are cached internally, and constructing equivalent Realm objects (for example, by using the same path or identifier) produces limited overhead.
+		try Realm(configuration: configuration)
+	}
 	
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		clearOld()
+
+		let realm = try! oneUseRealm()
+		clearOld(from: realm)
 		completion(nil)
 	}
 	
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
 		
-		clearOld()
-		
+		let realm = try! oneUseRealm()
+		clearOld(from: realm)
 		try! realm.write {
 			for image in feed {
 				realm.add(RealmFeedStoreCachedFeedImage.create(from: image, at: timestamp))
@@ -80,6 +88,7 @@ public final class RealmFeedStore: FeedStore {
 	
 	public func retrieve(completion: @escaping RetrievalCompletion) {
 
+		let realm = try! oneUseRealm()
 		guard let timestamp = realm.objects(RealmFeedStoreTimestamp.self).last?.timestamp else { return completion(.empty) }
 
 		let cached = realm.objects(RealmFeedStoreCachedFeedImage.self)
@@ -87,7 +96,7 @@ public final class RealmFeedStore: FeedStore {
 		completion(.found(feed: feedImages, timestamp: timestamp))
 	}
 	
-	private func clearOld() {
+	private func clearOld(from realm: Realm) {
 		try! realm.write {
 			realm.deleteAll()
 		}
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -66,9 +66,18 @@ public final class RealmFeedStore: FeedStore {
 		}
 	}
 	
+	private func writeToRealm(_ callback: (Realm)->()) throws {
+		try autoreleasepool {
+			let realm = try Realm(configuration: configuration)
+			try realm.write {
+				callback(realm)
+			}
+		}
+	}
+
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
 
-		try! accessRealm { realm in
+		try! writeToRealm { realm in
 			clearOld(from: realm)
 			completion(nil)
 		}
@@ -76,15 +85,14 @@ public final class RealmFeedStore: FeedStore {
 	
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
 		
-		try! accessRealm { realm in
+		try! writeToRealm { realm in
 			clearOld(from: realm)
-			try! realm.write {
-				for image in feed {
-					realm.add(RealmFeedStoreCachedFeedImage.create(from: image, at: timestamp))
-				}
-				
-				realm.add(RealmFeedStoreTimestamp.create(from: timestamp))
+			for image in feed {
+				realm.add(RealmFeedStoreCachedFeedImage.create(from: image, at: timestamp))
 			}
+			
+			realm.add(RealmFeedStoreTimestamp.create(from: timestamp))
+			
 			completion(nil)
 		}
 	}
@@ -101,8 +109,6 @@ public final class RealmFeedStore: FeedStore {
 	}
 	
 	private func clearOld(from realm: Realm) {
-		try! realm.write {
-			realm.deleteAll()
-		}
+		realm.deleteAll()
 	}
 }
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -57,7 +57,7 @@ public final class RealmFeedStore: FeedStore {
 		self.realm = try! Realm()
 	}
 	
-
+	
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
 		fatalError()
 	}
@@ -74,20 +74,11 @@ public final class RealmFeedStore: FeedStore {
 	}
 	
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		let cached = realm.objects(RealmFeedStoreCachedFeedImage.self)
-		if cached.count > 0 {
-			let feedImages = cached.map {
-				$0.localFeedImage
-			}
-			
-			let timestamp = realm.objects(RealmFeedStoreTimestamp.self).first!.timestamp!			
-			
-			completion(.found(feed: Array(feedImages), timestamp: timestamp))
-		}
-		else {
-			completion(.empty)
-		}
+
+		guard let timestamp = realm.objects(RealmFeedStoreTimestamp.self).first?.timestamp else { return completion(.empty) }
+
+		let cached = realm.objects(RealmFeedStoreCachedFeedImage.self)		
+		let feedImages = Array(cached.map(\.localFeedImage))
+		completion(.found(feed: feedImages, timestamp: timestamp))
 	}
-	
-	
 }
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -59,6 +59,7 @@ public final class RealmFeedStore: FeedStore {
 	
 	
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		clearOld()
 		completion(nil)
 	}
 	
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -75,13 +75,11 @@ public final class RealmFeedStore: FeedStore {
 		queue.async { [unowned self] in
 			do {
 				let realm = try Realm(configuration: self.configuration, queue: self.queue)
-//				try realm.write {
 				try ObjectiveCExceptions.performTry {
 					realm.beginWrite()
 				}
 				callback(.success(realm))
 				try realm.commitWrite()
-//				}
 			}
 			catch {
 				callback(.failure(error))
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -60,17 +60,21 @@ public final class RealmFeedStore: FeedStore {
 	}
 		
 	private func accessRealm(_ callback: (Realm)->()) throws {
-		try autoreleasepool {
-			let realm = try Realm(configuration: configuration)
-			callback(realm)
+		try queue.sync {
+			try autoreleasepool {
+				let realm = try Realm(configuration: configuration, queue: queue)
+				callback(realm)
+			}
 		}
 	}
 	
 	private func writeToRealm(_ callback: (Realm)->()) throws {
-		try autoreleasepool {
-			let realm = try Realm(configuration: configuration)
-			try realm.write {
-				callback(realm)
+		try queue.sync {
+			try autoreleasepool {
+				let realm = try Realm(configuration: configuration, queue: queue)
+				try realm.write {
+					callback(realm)
+				}
 			}
 		}
 	}
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -59,14 +59,14 @@ public final class RealmFeedStore: FeedStore {
 		self.queue = DispatchQueue(label: "\(type(of: Self.self))", qos: .userInitiated, autoreleaseFrequency: .workItem)
 	}
 	
-	private func accessRealm(_ callback: @escaping (Realm?)->()) throws {
+	private func accessRealm(_ callback: @escaping (Result<Realm, Error>)->()) {
 		queue.async {
 			do {
 				let realm = try Realm(configuration: self.configuration)
-				callback(realm)
+				callback(.success(realm))
 			}
 			catch {
-				
+				callback(.failure(error))
 			}
 		}
 	}
@@ -112,17 +112,19 @@ public final class RealmFeedStore: FeedStore {
 	}
 	
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-
-		try! accessRealm { realm in
-			guard let realm = realm else {
-				return completion(.empty)
+		
+		accessRealm { result in
+			switch result {
+			case .failure(let error):
+				return completion(.failure(error))
+			
+			case .success(let realm):
+				guard let timestamp = realm.objects(RealmFeedStoreTimestamp.self).last?.timestamp else { return completion(.empty) }
+				
+				let cached = realm.objects(RealmFeedStoreCachedFeedImage.self)
+				let feedImages = Array(cached.map(\.localFeedImage))
+				completion(.found(feed: feedImages, timestamp: timestamp))
 			}
-			
-			guard let timestamp = realm.objects(RealmFeedStoreTimestamp.self).last?.timestamp else { return completion(.empty) }
-			
-			let cached = realm.objects(RealmFeedStoreCachedFeedImage.self)
-			let feedImages = Array(cached.map(\.localFeedImage))
-			completion(.found(feed: feedImages, timestamp: timestamp))
 		}
 	}
 }
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -1,0 +1,28 @@
+//
+//  RealmFeedStore.swift
+//  FeedStoreChallenge
+//
+//  Created by Joseph Wardell on 12/31/20.
+//  Copyright © 2020 Essential Developer. All rights reserved.
+//
+
+import Foundation
+
+public final class RealmFeedStore: FeedStore {
+	
+	public init() {}
+	
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		fatalError()
+	}
+	
+	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+		fatalError()
+	}
+	
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		completion(.empty)
+	}
+	
+	
+}
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -53,8 +53,9 @@ final class RealmFeedStoreTimestamp: Object {
 public final class RealmFeedStore: FeedStore {
 	
 	let realm: Realm
-	public init() {
-		self.realm = try! Realm()
+	public init(_ fileURL: URL) {
+		let configuration = Realm.Configuration(fileURL: fileURL)
+		self.realm = try! Realm(configuration: configuration)
 	}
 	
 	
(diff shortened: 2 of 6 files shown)


Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+
+}
+
+extension FeedStoreChallengeTests: FailableInsertFeedStoreSpecs {
+
+	func test_insert_deliversErrorOnInsertionError() {
+
+		// trying to write into a realm that is readonly will cause an error
+		let sut = RealmFeedStore(testSpecificReadOnlyStoreURL(), readOnly: true)
+		trackForMemoryLeaks(sut)
```
caiozullo, line 178: Use the `makeSUT` method so you don't need to manually run the `trackForMemoryLeaks` method. Also, if there's a change in the initializer, you won't need to update many test methods.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+	func test_insert_hasNoSideEffectsOnInsertionError() {
+
+		// In order to test retrieval after a failed insert
+		// we need an existing realm file that the retrieve can read
+		// but we need it to be empty when insert() is called
+		writeEmptyRealmFile(at: testSpecificReadOnlyStoreURL())
+
+		// trying to insert into a realm that is readonly will cause an error
+		let sut = RealmFeedStore(testSpecificReadOnlyStoreURL(), readOnly: true)
+		trackForMemoryLeaks(sut)
```
caiozullo, line 194: Use the `makeSUT` method so you don't need to manually run the `trackForMemoryLeaks` method. Also, if there's a change in the initializer, you won't need to update many test methods.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		let exp1 = expectation(description: "Wait for dummy insertion")
+		existingRealm.insert(uniqueImageFeed(), timestamp: Date()) {_ in
+			exp1.fulfill()
+		}
+
+		let exp2 = expectation(description: "Wait for dummy deeltion")
+		existingRealm.deleteCachedFeed {_ in
+			exp2.fulfill()
+		}
+		wait(for: [exp1, exp2], timeout: 1.0)
```
caiozullo, line 216: Use the provided `insert` / `delete` test helpers to simplify this implementation.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+	
+}
+
+extension FeedStoreChallengeTests: FailableDeleteFeedStoreSpecs {
+
+	func test_delete_deliversErrorOnDeletionError() {
+
+		// trying to delete from a realm that is readonly will cause an error
+		let sut = RealmFeedStore(testSpecificReadOnlyStoreURL(), readOnly: true)
+		trackForMemoryLeaks(sut)
```
caiozullo, line 227: Use the `makeSUT` method so you don't need to manually run the `trackForMemoryLeaks` method. Also, if there's a change in the initializer, you won't need to update many test methods.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+	func test_delete_hasNoSideEffectsOnDeletionError() {
+
+		// In order to test retrieval after a failed deletion
+		// we need an existing realm file that the retrieve can read
+		// but we need it to be empty when insert() is called
+		writeEmptyRealmFile(at: testSpecificReadOnlyStoreURL())
+		
+		// trying to delete from a realm that is readonly will cause an error
+		let sut = RealmFeedStore(testSpecificReadOnlyStoreURL(), readOnly: true)
+		trackForMemoryLeaks(sut)
```
caiozullo, line 243: Use the `makeSUT` method so you don't need to manually run the `trackForMemoryLeaks` method. Also, if there's a change in the initializer, you won't need to update many test methods.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+			realmURL,
+			realmURL.appendingPathExtension("lock"),
+			realmURL.appendingPathExtension("note"),
+			realmURL.appendingPathExtension("management")
+		]
+		for URL in realmURLs {
+			do {
+				try FileManager.default.removeItem(at: URL)
+			} catch {
+				// handle error
```
caiozullo, line 301: The error should be handled, e.g., failed assertion. Otherwise, the environment could be in the wrong state when another test starts.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		clearRealmFiles(at: testSpecificStoreURL())
+	}
+	
+	private func clearRealmFiles(at realmURL: URL) {
+		let realmURLs = [
+			realmURL,
+			realmURL.appendingPathExtension("lock"),
+			realmURL.appendingPathExtension("note"),
+			realmURL.appendingPathExtension("management")
+		]
```
caiozullo, line 296: This looks risky - the URLs and extensions are private Realm details. 

What if Realm adds or removes path extensions in the future? You'd have to update it here every time. 

Is there a documented way of deleting a store? 

Maybe just use `deleteCache(from: makeSUT(...))` instead.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+	
+	private func testSpecificStoreURL() -> URL {
+		let out = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!.appendingPathComponent("\(type(of: self)).realm")
+		return out
+	}
+
+	private func testSpecificReadOnlyStoreURL() -> URL {
+		let out = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!.appendingPathComponent("\(type(of: self))_readonly.realm")
+		return out
+	}
```
caiozullo, line 279: Is it required to have a different readonly URL? 

Can't you use the same `testSpecificStoreURL` but with the read-only flag?

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		
+		// trying to delete from a realm that is readonly will cause an error
+		let sut = RealmFeedStore(testSpecificReadOnlyStoreURL(), readOnly: true)
+		trackForMemoryLeaks(sut)
+
+		print(testSpecificReadOnlyStoreURL())
+		
+		assertThatDeleteHasNoSideEffectsOnDeletionError(on: sut)
+		
+		clearRealmFiles(at: testSpecificReadOnlyStoreURL())
```
caiozullo, line 249: The realm files should be cleaned in `tearDown` already. You can remove this line - just make sure you use the same URL for all tests.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+
+	func test_delete_deliversErrorOnDeletionError() {
+
+		// trying to delete from a realm that is readonly will cause an error
+		let sut = RealmFeedStore(testSpecificReadOnlyStoreURL(), readOnly: true)
+		trackForMemoryLeaks(sut)
+
+		assertThatDeleteDeliversErrorOnDeletionError(on: sut)
+		
+		clearRealmFiles(at: testSpecificReadOnlyStoreURL())
```
caiozullo, line 231: The realm files should be cleaned in `tearDown` already. You can remove this line - just make sure you use the same URL for all tests.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		// but we need it to be empty when insert() is called
+		writeEmptyRealmFile(at: testSpecificReadOnlyStoreURL())
+
+		// trying to insert into a realm that is readonly will cause an error
+		let sut = RealmFeedStore(testSpecificReadOnlyStoreURL(), readOnly: true)
+		trackForMemoryLeaks(sut)
+
+		assertThatInsertHasNoSideEffectsOnInsertionError(on: sut)
+
+		clearRealmFiles(at: testSpecificReadOnlyStoreURL())
```
caiozullo, line 198: The realm files should be cleaned in `tearDown` already. You can remove this line - just make sure you use the same URL for all tests.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+
+	func test_insert_deliversErrorOnInsertionError() {
+
+		// trying to write into a realm that is readonly will cause an error
+		let sut = RealmFeedStore(testSpecificReadOnlyStoreURL(), readOnly: true)
+		trackForMemoryLeaks(sut)
+
+		assertThatInsertDeliversErrorOnInsertionError(on: sut)
+		
+		clearRealmFiles(at: testSpecificReadOnlyStoreURL())
```
caiozullo, line 182: The realm files should be cleaned in `tearDown` already. You can remove this line - just make sure you use the same URL for all tests.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		// then the realm will return an error when it tries to read
+		let sut = makeSUT(at: invalidStoreURL())
+
+		assertThatRetrieveDeliversFailureOnRetrievalError(on: sut)
+	}
+
+	func test_retrieve_hasNoSideEffectsOnFailure_forInvalidURL() {
+
+		// if the store url is invalid,
+		// then the realm will return an error when it tries to read
```
caiozullo, line 164: The scope is clear here - no need for comments. You can remove it.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		let sut = makeSUT(at: noPermissionsStoreURL())
+
+		assertThatRetrieveHasNoSideEffectsOnFailure(on: sut)
+	}
+
+	func test_retrieve_deliversFailureOnRetrievalError_forInvalidURL() {
+
+		// if the store url is invalid,
+		// then the realm will return an error when it tries to read
```
caiozullo, line 155: The scope is clear here - no need for comments. You can remove it.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		let sut = makeSUT(at: noPermissionsStoreURL())
+
+		assertThatRetrieveDeliversFailureOnRetrievalError(on: sut)
+	}
+
+	func test_retrieve_hasNoSideEffectsOnFailure_forPermissionsError() {
+
+		// if the current user doesn't have permission to read the store url passed in,
+		// then the realm will return an error when it tries to read
```
caiozullo, line 146: The scope is clear here - no need for comments. You can remove it.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		try! "not valid realm data".write(to: testSpecificStoreURL(), atomically: false, encoding: .utf8)
+		let sut = makeSUT(at: testSpecificStoreURL())
+
+		assertThatRetrieveHasNoSideEffectsOnFailure(on: sut)
+	}
+
+	func test_retrieve_deliversFailureOnRetrievalError_forPermissionsError() {
+		
+		// if the current user doesn't have permission to read the store url passed in,
+		// then the realm will return an error when it tries to read
```
caiozullo, line 137: The scope is clear here - no need for comments. You can remove it.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
-////		assertThatDeleteHasNoSideEffectsOnDeletionError(on: sut)
-//	}
-//
-//}
+		assertThatRetrieveDeliversFailureOnRetrievalError(on: sut)
+	}
+
+	func test_retrieve_hasNoSideEffectsOnFailure() {
+
+		// if there's already invalid data there, then the realm will return an error when it tries to read
```
caiozullo, line 127: The scope is clear here - no need for comments. You can remove it.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
-//	func test_insert_hasNoSideEffectsOnInsertionError() {
-////		let sut = makeSUT()
-////
-////		assertThatInsertHasNoSideEffectsOnInsertionError(on: sut)
-//	}
-//
-//}
+	func test_retrieve_deliversFailureOnRetrievalError() {
+		
+		// if there's already invalid data there, then the realm will return an error when it tries to read
```
caiozullo, line 118: The scope is clear here - no need for comments. You can remove it.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +27,90 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+		clearRealmFiles(at: testSpecificStoreURL())
 	}
 	
+	private func clearRealmFiles(at realmURL: URL) {
+		let realmURLs = [
+			realmURL,
+			realmURL.appendingPathExtension("lock"),
+			realmURL.appendingPathExtension("note"),
+			realmURL.appendingPathExtension("management")
+		]
```
caiozullo, line 106: This looks risky - the URLs and extensions are private Realm details. 

What if Realm adds or removes path extensions in the future? You'd have to update it here every time. 

Is there a documented way of deleting a store? 

Maybe just use `deleteCache(from: makeSUT(...))` instead.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+				callback(.success(realm))
+			}
+			catch {
+				callback(.failure(error))
+			}
+		}
+	}
+	
+	private func writeToRealm(_ callback: @escaping (Result<Realm, Error>)->()) {
+		queue.async { [unowned self] in
```
caiozullo, line 75: An `unowned` reference can lead to runtime errors here. Use `weak` or strong.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+	}
+		
+	fileprivate var localFeedImage: LocalFeedImage {
+		LocalFeedImage(id: UUID(uuidString: id!)!, description: desc, location: location, url: URL(string: url!)!)
+	}
+}
+
+/// This is an internal type to RealmFeedStore
+/// It should not be used by any code outside this file
+// we would make it fileprivate, but Realm complains with an error
```
caiozullo, line 39: This class is internal by default. Internal classes cannot be used by other client modules, so there's no problem here - they don't need to be private. You can delete the comment.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+//  Created by Joseph Wardell on 12/31/20.
+//  Copyright © 2020 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+/// This is an internal type to RealmFeedStore
+/// It should not be used by any code outside this file
+// we would make it fileprivate, but Realm complains with an error
```
caiozullo, line 14: This class is internal by default. Internal classes cannot be used by other client modules, so there's no problem here - they don't need to be private. You can delete the comment.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+	let queue: DispatchQueue
+	public init(_ fileURL: URL, readOnly: Bool = false) {
+		self.configuration = Realm.Configuration(fileURL: fileURL, readOnly: readOnly)
+		self.queue = DispatchQueue(label: "\(type(of: Self.self))", qos: .userInitiated, autoreleaseFrequency: .workItem)
+	}
+	
+	private func accessRealm(_ callback: @escaping (Result<Realm, Error>)->()) {
+		queue.async {
+			do {
+				let realm = try Realm(configuration: self.configuration)
```
caiozullo, line 65: Should you also pass the queue here as you did in the `writeToRealm` method?

```
let realm = try Realm(configuration: self.configuration, queue: self.queue)
```

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,204 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+
+}
+
+// MARK:- Helpers
+
+extension FeedStoreChallengeTests {
+	
+	private func makeSUT(at fileURL: URL? = nil, file: StaticString = #filePath, line: UInt = #line) -> FeedStore {
+		let fileURL = fileURL ?? testSpecificStoreURL()
+		let sut = RealmFeedStore(fileURL)
```
caiozullo, line 260: You could try running those tests with an in-memory Realm store to speed up the process. You can still use the real URLs in the Integration Tests.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+/// This is an internal type to RealmFeedStore
+/// It should not be used by any code outside this file
+// we would make it fileprivate, but Realm complains with an error
+final class RealmFeedStoreCachedFeedImage: Object {
+	@objc fileprivate dynamic var id: String?
+	@objc fileprivate dynamic var desc: String?
+	@objc fileprivate dynamic var location: String?
+	@objc fileprivate dynamic var url: String?
+
+	fileprivate class func create(from localFeedImage: LocalFeedImage, at timestamp: Date) -> RealmFeedStoreCachedFeedImage {
```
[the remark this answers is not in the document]
jaywardell: Hi Caio,
Thanks for the voluminous and thorough feedback.  If only for my later reference, I'm going to reply to each comment as I tackle it, but I won't push again until I've addressed every issue you bring up. Please excuse the verbosity.

for this comment: I didn't think it through. I assumed that it was unsafe to override a Realm object initializer, so I assumed I needed these to be optional in order to ensure that all values were set upon init (to avoid the error: class "RealmFeedStoreCachedFeedImage" has no initializers).  It didn't occur to me to go ahead and try it and let the tests tell me if it was okay.

id and url are now non optional as you suggest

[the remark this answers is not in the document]
jaywardell: Hi Caio,
Thanks for the voluminous and thorough feedback. 

If only for my later reference, I'm going to reply to each comment as I tackle it, but I won't push again until I've addressed every issue you bring up. Please excuse the verbosity.

for this comment: This was a vestigial remnant from an earlier approach.  I'm embarrassed I forgot to remove it.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+//
+
+import Foundation
+import RealmSwift
+
+/// This is an internal type to RealmFeedStore
+/// It should not be used by any code outside this file
+// we would make it fileprivate, but Realm complains with an error
+final class RealmFeedStoreCachedFeedImage: Object {
+	@objc fileprivate dynamic var id: String?
```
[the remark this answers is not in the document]
jaywardell: I didn't think this through, I'll admit. I assumed that it was unsafe to override a Realm object initializer, so I assumed I needed these to be optional in order to ensure that all values were set upon init (to avoid the error: class "RealmFeedStoreCachedFeedImage" has no initializers). It didn't occur to me to go ahead and try it and let the tests tell me if it was okay.

id and url are now non optional as you suggest

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+	@objc fileprivate dynamic var desc: String?
+	@objc fileprivate dynamic var location: String?
+	@objc fileprivate dynamic var url: String?
```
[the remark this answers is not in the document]
jaywardell: done

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+		out.id = localFeedImage.id.uuidString
+		out.desc = localFeedImage.description
+		out.location = localFeedImage.location
+		out.url = localFeedImage.url.absoluteString
+				
+		return out
+	}
+		
+	fileprivate var localFeedImage: LocalFeedImage {
+		LocalFeedImage(id: UUID(uuidString: id!)!, description: desc, location: location, url: URL(string: url!)!)
```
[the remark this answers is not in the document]
jaywardell: agreed.  I think I was worried about the retrieve() method delivering a list of [LocalFeedImage] shorter than what had been inserted.  Of course, then the tests would have failed.  So again, I failed to trust the tests.

--fixed

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+		out.timestamp = timestamp
+		
+		return out
+	}
+}
+
+// MARK:-
+public final class RealmFeedStore: FeedStore {
+	
+	let configuration: Realm.Configuration
```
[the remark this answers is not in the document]
jaywardell: oops. done

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+	let queue: DispatchQueue
```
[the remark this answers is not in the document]
jaywardell: oops. done

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+	fileprivate var localFeedImage: LocalFeedImage {
+		LocalFeedImage(id: UUID(uuidString: id!)!, description: desc, location: location, url: URL(string: url!)!)
+	}
+}
+
+/// This is an internal type to RealmFeedStore
+/// It should not be used by any code outside this file
+// we would make it fileprivate, but Realm complains with an error
+final class RealmFeedStoreTimestamp: Object {
+	@objc fileprivate dynamic var timestamp: Date?
```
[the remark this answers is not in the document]
jaywardell: same weird logic as for RealmFeedStoreCachedFeedImage
fixed

Tests/FeedStoreChallengeTests.swift
jaywardell: I considered that but chose against it for two reasons: 
1) I felt like it was cheating.  As I think about it, the integration tests should be sufficient to verify disk access, but running tests on an in-Memory instance felt somehow inauthentic. 
 I think I'm still shaky on the distinction between unit tests and integration tests.  When it's explained to me, I get it, but as I'm working I find the line to be much fuzzier.

2) Since I was creating a new Realm instance on each call to accessRealm() or writeToRealm(), an in-Memory Realm store would fail almost any retrieval test automatically, and simultaneous writes from different read-only and read-write Realms would cause runtime errors.  On retrospect, this was an obvious signal that internal state was bleeding out into the tests, and I should have handled it right away.  When I started having one Realm instance per RealmFeedStore, these issues went away.

Thanks for this recommendation.  It forced me to rethink more of my implementation than I expected and it showed me how brittle my implementation was.

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: yes, I should have been doing that.  I wasn't encountering problems until I tried using an in-Memory Realm store per your suggestion below, then things started breaking.

Thanks

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: ok.  I was thinking about hypothetical future me working on this particular module, but yes, that's a very low-percentage possibility, admittedly

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: thanks, done
(see other note...)

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: another one I should have caught. Thanks, done

Tests/FeedStoreChallengeTests.swift
jaywardell: done

Tests/FeedStoreChallengeTests.swift
jaywardell: done

Tests/FeedStoreChallengeTests.swift
jaywardell: done

Tests/FeedStoreChallengeTests.swift
jaywardell: done

Tests/FeedStoreChallengeTests.swift
jaywardell: done

Tests/FeedStoreChallengeTests.swift
jaywardell: done

Tests/FeedStoreChallengeTests.swift
jaywardell: It WAS necessary when I was using file-based Realm stores for all these tests, since parallel tests on the same Realm file, some of them readonly and others read-write, caused Realm to throw errors. 

Since I've moved most tests to using an in-memory Realm feed store per your recommendation, there's no issue.
done

Tests/FeedStoreChallengeTests.swift
jaywardell: okay, done

Tests/FeedStoreChallengeTests.swift
jaywardell: okay, done

Tests/FeedStoreChallengeTests.swift
jaywardell: okay, done

Tests/FeedStoreChallengeTests.swift
jaywardell: okay, done

FeedStoreChallenge/RealmFeedStore.swift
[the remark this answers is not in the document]
jaywardell: Why did I assume that it's unsafe to override a Realm initializer? all the example code, including Realm's, never does. They always initialize a blank object and then fill in parameters afterward. So I assumed that I needed to follow the same pattern.  I should have tested it before making the assumption.

Tests/FeedStoreChallengeTests.swift
jaywardell: yes, that's much simpler.  Thanks.
done

Tests/FeedStoreChallengeTests.swift
jaywardell: Yes, this was necessary only because I was using different files for different tests.  
Thanks,
fixed

Tests/FeedStoreChallengeTests.swift
jaywardell: Yes, this was necessary only because I was using different files for different tests.  
Thanks,
fixed

Tests/FeedStoreChallengeTests.swift
jaywardell: Yes, this was necessary only because I was using different files for different tests.  
Thanks,
fixed

Tests/FeedStoreChallengeTests.swift
jaywardell: Yes, this was necessary only because I was using different files for different tests.  
Thanks,
fixed

Tests/FeedStoreChallengeTests.swift
jaywardell: This code came directly from Realm's documentation at https://realm.io/docs/swift/latest/#migrations subsection "Deleting Realm files". So yes, there's a documented way of deleting a realm store, and it's this one :-) Granted, the documentation talks about using an autoreleasepool, but this code is called only at the beginning or the end of a test, so I don't think it's an issue.

I tried using deleteCache(from: makeSUT(...)) and found that it caused synchronization issues between tests. Essentially, one Realm instance is trying to init from a file as another is deleting it.

So right now I think this is the best approach, imho.

Tests/FeedStoreChallengeTests.swift
jaywardell: I hadn't considered that situation. Thanks.  
fixed

FeedStoreChallenge/RealmFeedStore.swift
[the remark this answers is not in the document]
jaywardell: and by that I mean that I was using weird logic at the time...

Tests/FeedStoreIntegrationTests.swift
jaywardell: see my response to the other comment about this code in FeedStoreChallengeTests.swift

Tests/FeedStoreChallengeTests.swift
jaywardell: Of course, all that said, there's a static method deleteFiles(for: Configuration) on Realm that appears to do the same thing and would be updated by Realm should they ever change things.  So I switched to using that.

Thanks for making me think about this.

Tests/FeedStoreChallengeTests.swift
jaywardell: As I suspected, once I did this, a lot of your other suggestions became much easier to handle.  This really simplified the testing and forced me to shore up the implementation in ways I hadn't considered. Thanks for the suggestion.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,140 @@
...
+		
+	fileprivate var localFeedImage: LocalFeedImage {
+		LocalFeedImage(id: UUID(uuidString: id!)!, description: desc, location: location, url: URL(string: url!)!)
+	}
+}
+
+/// This is an internal type to RealmFeedStore
+/// It should not be used by any code outside this file
+// we would make it fileprivate, but Realm complains with an error
+final class RealmFeedStoreTimestamp: Object {
```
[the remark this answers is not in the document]
jaywardell: yes, that does make the insert() and deleteCache() methods simpler. Thanks.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,168 @@
...
+
+	public init(fileURL: URL, readOnly: Bool = false) {
+		self.configuration = Realm.Configuration(fileURL: fileURL, readOnly: readOnly)
+		self.queue = DispatchQueue(label: "\(type(of: Self.self))", qos: .userInitiated, autoreleaseFrequency: .workItem)
+	}
+	
+	public init(inMemoryIdentifier: String, readOnly: Bool = false) {
+		self.configuration = Realm.Configuration(inMemoryIdentifier: inMemoryIdentifier, readOnly: readOnly)
+		self.queue = DispatchQueue(label: "\(type(of: Self.self))", qos: .userInitiated, autoreleaseFrequency: .workItem)
+	}
```
caiozullo, line 78: There's duplication in these two initializers. Maybe it should be a single initializer:

```
public init(configuration: Realm.Configuration) 
```

So the clients can set any configuration they want, e.g., creating a read-only in-memory store.

This way it doesn't leak test config details into prod like read-only which is for testing purposes only.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,168 @@
...
+		guard canWrite else { return callback(.failure(WriteError.readonlyStore)) }
+
+		queue.async { [weak self] in
+			do {
+				if let realm = try self?.getRealm() {
+					try realm.write {
+						callback(.success(realm))
+					}
+				}
+			}
```
caiozullo, line 112: Why does `self` need to be weak here but not in the `accessRealm` method?

I believe it'd be best to capture a strong reference to self for this operation if possible.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +111,163 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+
+		assertThatInsertHasNoSideEffectsOnInsertionError(on: sut)
+	}
+
+	private func writeEmptyRealmFile(at fileURL: URL) {
+		
+		// create an existing realm at the standard test location
+		// and write to it, but leave it empty
+		// so that a file exists at the expected path
+		let existingRealm = RealmFeedStore(fileURL: fileURL)
```
caiozullo, line 189: I believe you can use the `makeSUT` method here.

FeedStoreChallenge.xcodeproj/xcshareddata/xcschemes/FeedStoreChallenge.xcscheme
```diff
@@ -50,6 +50,7 @@
       <Testables>
          <TestableReference
             skipped = "NO"
+            parallelizable = "YES"
```
caiozullo, line 53: Please, revert changes to the scheme - this is not part of the challenge.

BridgingHeader.h
```diff
@@ -0,0 +1,24 @@
+//
+//  BridgingHeader.h
```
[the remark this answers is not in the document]
jaywardell: done

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,168 @@
...
+	fileprivate class func create(from localFeedImage: LocalFeedImage) -> RealmFeedStoreCachedFeedImage {
+		
+		let out = RealmFeedStoreCachedFeedImage()
+		out.id = localFeedImage.id.uuidString
+		out.desc = localFeedImage.description
+		out.location = localFeedImage.location
+		out.url = localFeedImage.url.absoluteString
+				
+		return out
+	}
```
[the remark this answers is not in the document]
jaywardell: Actually, I was able to even go to using convenience initializers for these.  Thanks for pointing this out. The create() methods were definitely no longer needed

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,168 @@
...
+		timestamp = Date.distantPast
+		super.init()
+	}
+	
+	static func create(at timestamp: Date, with feedImages: [LocalFeedImage]) -> RealmFeedCache {
+		let out = RealmFeedCache()
+		out.timestamp = timestamp
+		out.images.append(objectsIn: feedImages.map(RealmFeedStoreCachedFeedImage.create(from:)))
+		return out
+	}
```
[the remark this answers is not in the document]
jaywardell: as above, thanks

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: I see what you're saying about DRY here, and about leaking testing details into the production code.

I actually went a little farther.  I've only tested this code against a limited number of possible Realm.Configurations: inMemory, readonly and fileURL.  There's no way to know for sure how other Realm.Configurations would affect the code.  For this reason, I don't think I would want client code to be able to pass just any configuration into the init.

So instead, I created 2 initializers: 1 that only accepts those possible configuration options (fileURL or inMemory, readONly or readwrite) and is marked as "for testing only" in the comments, and one that accepts only a file URL and is commented as "for production code."

I considered making the "for-testing only" initializer internal and using @testable import in the test file, but chose against it.

and btw, the integration tests end up using the "for production code" init without me having to do anything.  I only realized after the fact.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,168 @@
...
+				
+			case .success(let realm):
+				realm.deleteAll()
+
+				let cache = RealmFeedCache.create(at: timestamp, with: feed)
+				realm.add(cache)
+				
+				completion(nil)
+			}
+		}
```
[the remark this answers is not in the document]
jaywardell: or I could just get rid of the accessRealm and writeRealm methods.  That makes things simpler to read and accomplishes the same thing

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: done, given that I got rid of accessRealm and writeToRealm anyway.

almost every closure I've ever written that references self, I've done the weak self dance because I was never sure if it would cause a retain cycle.  Having the tests in place to verify that I'm not retaining self is really empowering.  It gives me a lot of confidence.  Awesome.

Tests/FeedStoreChallengeTests.swift
jaywardell: true, but that would mean using the makeSUT() method to do something other than making a system under test, right? The purpose of the method is to set up state to cause a failure, not to make an assertion.

 I made the change, but I'm not sure it makes the meaning of the code clearer...

FeedStoreChallenge.xcodeproj/xcshareddata/xcschemes/FeedStoreChallenge.xcscheme
jaywardell: oops.  just curiosity made me turn it on. fixed

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,163 @@
...
+import Foundation
+import RealmSwift
+
+final class RealmFeedStoreCachedFeedImage: Object {
+	@objc fileprivate dynamic var id: String
+	@objc fileprivate dynamic var desc: String?
+	@objc fileprivate dynamic var location: String?
+	@objc fileprivate dynamic var url: String
+
+	private override init() {
```
caiozullo, line 18: I believe you can remove this initializer and have only a single one.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,163 @@
...
+		self.id = ""
+		self.url = ""
+		super.init()
+	}
+
+	convenience init(localFeedImage: LocalFeedImage) {
```
caiozullo, line 24: If you remove the `convenience` keyword, this could be the single initializer for this class.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,163 @@
...
+	
+	@objc fileprivate dynamic var timestamp: Date
+	fileprivate dynamic var images = List<RealmFeedStoreCachedFeedImage>()
+	
+	fileprivate override init() {
+		timestamp = Date.distantPast
+		super.init()
+	}
+	
+	convenience init(timestamp: Date, feed: [LocalFeedImage]) {
```
caiozullo, line 51: If you remove the `convenience` keyword, this could be the single initializer for this class.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,163 @@
...
+	enum WriteError: Error { case readonlyStore }
+
+	public enum Configuration {
+		case fileURL(URL)
+		case inMemoryIdentifier(String)
+	}
+	
+	/// use this initializer for testing purposes only
+	///
+	/// use init(fileURL: URL) for production code
```
caiozullo, line 73: This initializer can also be used in prod, for example, if you want a temporary in-memory store. You can remove the comment.

Otherwise, I'd advise you to find solutions that don't require test-specific code in prod.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,163 @@
...
+	
+	private var _realm: Realm?
+	private func getRealm() throws -> Realm {
+		if let existing = _realm { return existing }
+		_realm = try Realm(configuration: self.configuration, queue: queue)
+		return _realm!
+	}
+	
+	private var canWrite: Bool { !configuration.readOnly }
+	private func ensureSafeConfiguration() -> Error? {
```
caiozullo, line 97: It's more idiomatic to throw errors instead of returning optional errors in Swift.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,163 @@
...
+		// calling write on a readonly Realm store causes an Objective-C exception to be thrown
+		// so let's just throw our own error before it happens
+		if !canWrite {
+			return WriteError.readonlyStore
+		}
+		return nil
+	}
+	
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		if let error = ensureSafeConfiguration() { return completion(error) }
```
caiozullo, line 107: Making the store read-only (a test-specific detail) is over-complicating the implementation with the guard statements. It'd be best to find another solution where the test details stay on the test side.

caiozullo (commented):
Hi @jaywardell,

- All tests are passing ✅ 

I left an important review comment. Check it out and feel free to push changes.

Great job!

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: I tried and couldn't get it to work.

RealmFeedStoreCachedFeedImage is a subclass of RLMObjectBase, which has a single designated initializer: 

> - (instancetype)init NS_DESIGNATED_INITIALIZER;
> RLMObjectBase.h:32

eliminating the init at line 18 causes a runtime error:

> 2021-01-06 11:33:41.673273-0500 xctest[19619:4881238] FeedStoreChallenge/RealmFeedStore.swift: 12: 13: Fatal error: Use of unimplemented initializer 'init()' for class 'FeedStoreChallenge.RealmFeedStoreCachedFeedImage'

I've tried various combinations of calls to super.init(), giving default values to the properties, etc., Sometimes I get a different runtime error, or a compiler error about unimplemented init(), but this is the only approach I can get to work.  

Any suggestions?

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: see note from above

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: see note from above

FeedStoreChallenge/RealmFeedStore.swift
caiozullo: You need to define default values for all properties:

```
@objc fileprivate dynamic var id: String = ""
@objc fileprivate dynamic var url: String = ""
```

jaywardell: Screen Shot 2021-01-06 at 11 54 44 AM

> 2021-01-06 11:53:52.974897-0500 xctest[19796:4893428] FeedStoreChallenge/RealmFeedStore.swift: 12: 13: Fatal error: Use of unimplemented initializer 'init()' for class 'FeedStoreChallenge.RealmFeedStoreCachedFeedImage'

jaywardell: fuller content:

```
final class RealmFeedStoreCachedFeedImage: Object {
	@objc fileprivate dynamic var id: String = ""
	@objc fileprivate dynamic var desc: String?
	@objc fileprivate dynamic var location: String?
	@objc fileprivate dynamic var url: String = ""

//	private override init() {
//		self.id = ""
//		self.url = ""
//		super.init()
//	}

//	convenience
	init(localFeedImage: LocalFeedImage) {
		super.init()
		self.id = localFeedImage.id.uuidString
		self.desc = localFeedImage.description
		self.location = localFeedImage.location
		self.url = localFeedImage.url.absoluteString
//		super.init()
	}

```

caiozullo: You can check how it was done in other challenges, like [this one](https://github.com/essentialdevelopercom/ios-lead-essentials-feed-store-challenge/blob/a32ed8ad94ae825162fcd69ffce040f798eaa56c/FeedStoreChallenge/Cache.swift#L36).

jaywardell: Wait, I know I tested that combination (default values for non optional properties and calling self.init() rather than super.init()).  But now it works.  😣
Thanks.
Done

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: per the change above and the example you showed me, I still need the convenience keyword, but I've reduced the class to one init, at least

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: I think that brings me back to needing to catch NSException.  Or I could eliminate the test against in-Memory stores, but I se that as a regression, since it's an authentic way to make the failing tests fail.

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: I considered this.  Throwing from this method would require either 2 do/catch blocks everywhere it's called (one for calling this method and one for the actual work, as it happens on a different queue), or entering a separate queue just to then fail this test.

I consider this to be more or less a precondition at the call site:
`		if let error = ensureSafeConfiguration() { return completion(error) }
`
It it can't work, then throw right away.

It's less idiomatic, but I think it fits this use case better.

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: How about making it internal and then importing the framework as testable in the unit tests target? That way it's there for tests, but production code can only access it through the safer init(fileURL:)

btw, I added an init(inMemoryIdentifier:) convenience init for the case you mention, but now I'm wondering: All the happy-path tests against this class are using the  fileURL configuration (makeSUT(readonly: false)), so yes it may be useful for client code, but is it really tested?

jaywardell: Anyway, I'm pushing what I have.  Feel free to comment at your leisure. You've passed me, so I have access to several more hours of lessons, and I have a new client (hopefully).  So my plate is full.

Thanks again.   I'm learning a whole lot.

jaywardell: oh, and btw, this got me to take the leap and publish a tutorial on YouTube.  I've been thinking about this "Xcode Moments" series for a while.  This was an ideal jumping off point.  I hope you don't mind me quickly scrolling through my solution in public in the video...

FeedStoreChallenge/RealmFeedStore.swift
caiozullo: Yes, the idea was to have a single do/catch where you can call the `ensureSafeConfiguration`. But that would require running it inside the queue block.

The problem is that this class allows an invalid configuration (readonly) - The interface says you can insert/delete, but internally it might refuse the operation because of the readonly state. This class shouldn't allow read-only configurations if it cannot perform all its operations with it.

The readonly config helps with tests but complicates the implementation. it's a trade-off as always 🙂

FeedStoreChallenge/RealmFeedStore.swift
caiozullo: See my last comment.

FeedStoreChallenge/RealmFeedStore.swift
caiozullo: Feel free to record the video! And don't forget to share it in the Slack community 🙂

FeedStoreChallenge/RealmFeedStore.swift
jaywardell: with the new init structure, the readonly option is only available to test code.  Production code can't access it. So hopefully the tradeofff leans toward the positive...

caiozullo: Looks good!

PR #70, opened 2021-01-17
Realm Implementation - Jorge Lucena

jorgifumi:
Hi Caio & Mike!

This is my RealmFeedStore implementation. I'm not sure if it's finished since there are two tests I'm unable to make it pass 😟: `test_insert_hasNoSideEffectsOnInsertionError` and `test_delete_hasNoSideEffectsOnDeletionError`.

I choose Realm because I have less experience with it than with CoreData and I think it is a good chance to try it.

I implemented the `setup()` and `tearDown()` methods in the `FeedStoreChallengeTests` before seeing the `FeedStoreIntegrationTests` so I'm not sure if it is correct or I would have used an In-Memory configuration for the former.

Looking forward to your feedback.

Thanks!!

### DIFF
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -32,6 +32,7 @@ extension RealmFeedStore: FeedStore {
 		realmFeed.append(objectsIn: feed.map(RealmFeedImage.init))
 		let cache = Cache(value: [realmFeed, timestamp])
 		try! realm.write {
+			realm.deleteAll()
 			realm.add(cache)
 			completion(nil)
 		}
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -7,8 +7,11 @@
 //
 
 import Foundation
+import RealmSwift
 
 public final class RealmFeedStore {
+
+	let realm = try! Realm()
 	
 	public init() {
 		
@@ -17,14 +20,68 @@ public final class RealmFeedStore {
 
 extension RealmFeedStore: FeedStore {
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		
+		try! realm.write {
+			realm.deleteAll()
+			completion(nil)
+		}
 	}
 	
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
-		
+		let realmFeed = List<RealmFeedImage>()
+		realmFeed.append(objectsIn: feed.map(RealmFeedImage.init))
+		let cache = Cache(value: [realmFeed, timestamp])
+		try! realm.write {
+			realm.add(cache)
+			completion(nil)
+		}
 	}
 	
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		completion(.empty)
+		
+		let cache = realm.objects(Cache.self)
+		
+		if cache.isEmpty {
+			completion(.empty)
+		} else {
+			completion(.found(feed: cache[0].localFeed, timestamp: cache[0].timestamp))
+		}
+	}
+}
+
+class Cache: Object {
+	let feed = List<RealmFeedImage>()
+	@objc dynamic var timestamp = Date()
+	
+	var localFeed: [LocalFeedImage] {
+		return feed.compactMap { $0.local }
+	}
+}
+
+class RealmFeedImage: Object {
+	@objc dynamic var id: String = ""
+	@objc dynamic var imageDescription: String? = nil
+	@objc dynamic var location: String? = nil
+	@objc dynamic var urlString: String = ""
+	
+	convenience init(_ image: LocalFeedImage) {
+		self.init()
+		self.id = image.id.uuidString
+		self.imageDescription = image.description
+		self.location = image.location
+		self.urlString = image.url.absoluteString
+	}
+	
+	override class func primaryKey() -> String? {
+		"id"
+	}
+	
+	var local: LocalFeedImage? {
+		
+		guard let id = UUID(uuidString: id),
+			  let url = URL(string: urlString) else {
+			return nil
+		}
+		
+		return LocalFeedImage(id: id, description: imageDescription, location: location, url: url)
 	}
 }
--- a/FeedStoreChallenge/RealmFeedStore.swift
+++ b/FeedStoreChallenge/RealmFeedStore.swift
@@ -16,15 +16,20 @@ public final class RealmFeedStore {
 	public init(storeURL: URL) {
 		configuration = Realm.Configuration(fileURL: storeURL)
 	}
+
+	private var _realm: Realm?
+
+	private func getRealm() throws -> Realm {
+		try _realm ?? Realm(configuration: configuration)
+	}
 }
 
 extension RealmFeedStore: FeedStore {
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		let realm: Realm?
 		do {
-			realm = try Realm(configuration: configuration)
-			try! realm?.write {
-				realm?.deleteAll()
+			let realm = try getRealm()
+			try realm.write {
+				realm.deleteAll()
 				completion(nil)
 			}
 		} catch let error {
@@ -37,12 +42,11 @@ extension RealmFeedStore: FeedStore {
 		realmFeed.append(objectsIn: feed.map(RealmFeedImage.init))
 		let cache = Cache(value: [realmFeed, timestamp])
 
-		let realm: Realm?
 		do {
-			realm = try Realm(configuration: configuration)
-			try! realm?.write {
-				realm?.deleteAll()
-				realm?.add(cache)
+			let realm = try getRealm()
+			try realm.write {
+				realm.deleteAll()
+				realm.add(cache)
 				completion(nil)
 			}
 		} catch let error {
@@ -51,12 +55,11 @@ extension RealmFeedStore: FeedStore {
 	}
 	
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		let realm: Realm?
 		do {
-			realm = try Realm(configuration: configuration)
-			let cache = realm?.objects(Cache.self)
+			let realm = try getRealm()
+			let cache = realm.objects(Cache.self)
 
-			if let cache = cache?.first {
+			if let cache = cache.first {
 				completion(.found(feed: cache.localFeed, timestamp: cache.timestamp))
 			} else {
 				completion(.empty)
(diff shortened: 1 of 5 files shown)


FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,125 @@
...
+//
+
+import Foundation
+import RealmSwift
+
+public final class RealmFeedStore {
+
+	private let configuration: Realm.Configuration
+
+	private var _realm: Realm?
```
caiozullo, line 16: This property is never set. You can remove it.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+extension RealmFeedStore: FeedStore {
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		queue.async { [weak self] in
+			do {
+				guard let realm = try self?.getRealm() else {
+					return completion(nil)
+				}
```
caiozullo, line 35: There's no need to weakify `self` in this context - so you don't need to deal with optionals.

Moreover, if it cannot get the Realm instance, it shouldn't complete with `nil` - it should complete with an error since the `delete operation` failed.

Capturing a strong reference to `self` will solve this problem since there won't be an `Optional` to unwrap.

Or even better, you don't even need to capture `self` - for example:

```suggestion
		queue.async { [configuration, queue] in
			do {
				let realm = try Realm(configuration: configuration, queue: queue)
```

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+	
+	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+
+		queue.async { [weak self] in
+			do {
+				guard let realm = try self?.getRealm() else {
+					return completion(nil)
+				}
```
caiozullo, line 52: Same here - there's no need to weakify `self` in this context - see my other comment.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,125 @@
...
+			}
+		}
+	}
+	
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		queue.async { [weak self] in
+			do {
+				guard let realm = try self?.getRealm() else {
+					return completion(.empty)
+				}
```
caiozullo, line 69: Same here - there's no need to weakify `self` in this context - see my other comment.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +148,64 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+	}
+
+	func test_delete_hasNoSideEffectsOnDeletionError() {
+//		let noDeletionPermissionURL = cachesDirectory()
+//		let sut = makeSUT(storeURL: noDeletionPermissionURL)
 //
-//}
+//		assertThatDeleteHasNoSideEffectsOnDeletionError(on: sut)
+	}
+}
```
caiozullo, line 211: It's not possible to simulate a deletion error using the `cachesDirectory` as the store URL because it cannot even open the DB connection with it - I recommend you delete those tests since it's not actually testing what it's supposed to.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +148,64 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
 //
-//	func test_delete_hasNoSideEffectsOnDeletionError() {
-////		let sut = makeSUT()
-////
-////		assertThatDeleteHasNoSideEffectsOnDeletionError(on: sut)
-//	}
+//		assertThatInsertHasNoSideEffectsOnInsertionError(on: sut)
+	}
+
+}
```
caiozullo, line 194: It's not possible to simulate an insertion error using the `cachesDirectory` as the store URL because it cannot even open the DB connection with it - I recommend you delete those tests since it's not actually testing what it's supposed to.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,125 @@
...
+
+		queue.async { [weak self] in
+			do {
+				guard let realm = try self?.getRealm() else {
+					return completion(nil)
+				}
+				try realm.write {
+					realm.deleteAll()
+					realm.add(RealmFeedStore.map(feed, timestamp: timestamp))
+					completion(nil)
```
caiozullo, line 56: The `completion` should be outside the `realm.write` block. Otherwise, this method may call `completion` twice if the transaction fails (once with `nil` and another with the `error` caught in the `catch` block!

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,125 @@
...
+extension RealmFeedStore: FeedStore {
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		queue.async { [weak self] in
+			do {
+				guard let realm = try self?.getRealm() else {
+					return completion(nil)
+				}
+				try realm.write {
+					realm.deleteAll()
+					completion(nil)
```
caiozullo, line 38: The `completion` should be outside the `realm.write` block. Otherwise, this method may call `completion` twice if the transaction fails (once with `nil` and another with the `error` caught in the `catch` block!

Tests/FeedStoreChallengeTests.swift
```diff
@@ -19,84 +19,125 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+
+		assertThatSideEffectsRunSerially(on: sut)
 	}
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT(storeURL: URL? = nil, file: StaticString = #filePath, line: UInt = #line) -> FeedStore {
+		let sut = RealmFeedStore(storeURL: storeURL ?? testSpecificStoreURL())
```
caiozullo, line 109: You could run those tests with an in-memory Realm store to speed up the process. You can still use the real URLs in the Integration Tests.

caiozullo (commented):
Hi @jorgifumi, 

Thank you! 🙂

FeedStoreChallenge/RealmFeedStore.swift
jorgifumi: Thanks!

FeedStoreChallenge/RealmFeedStore.swift
jorgifumi: Good advice, thanks!

FeedStoreChallenge/RealmFeedStore.swift
jorgifumi: Done, thanks

Tests/FeedStoreChallengeTests.swift
jorgifumi: Ok, there is no other option to simulate deletion error on Realm then?

Tests/FeedStoreChallengeTests.swift
jorgifumi: I tried it but then all three `test_retrieve_deliversFoundValuesOnNonEmptyCache`, `test_retrieve_hasNoSideEffectsOnNonEmptyCache`, and `test_insert_overridesPreviouslyInsertedCacheValues` fails, I guess that is because I'm not capturing the Realm instance and it creates a new one in each operation, but since the initializer can throw it becomes tricky to solve all the problems at the same time... any advice on this?

jorgifumi:
Hi @caiozullo!

I resolved some of the comments but left a reply on others that I'm not sure how to act. Thanks in advance for your good advice.

Tests/FeedStoreChallengeTests.swift
caiozullo: It should work with multiple instances as long as you use the same `inMemoryIdentifier` and you hold a strong reference to at least one `Realm` instance using that identifier. 

From the docs:

*"Notice: When all in-memory Realm instances with a particular identifier go out of scope with no references, all data is freed for that Realm. It is recommended that you hold onto a strong reference to any created in-memory Realms for the duration of your app."*-https://realm.io/docs/swift/latest/#in-memory-realms

So instead of passing the `storeURL` in the initializer, you could pass a `Realm.Configuration` with either the `URL` (for integration tests) or an `inMemoryIdentifier` (for isolated tests). When using an in-memory id, make sure to create and hold a `Realm` instance with that id on the test side.

Tests/FeedStoreChallengeTests.swift
caiozullo: You can check how other students did it - either creating extra abstractions that can be mocked/stubbed or using `readonly` stores.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -19,84 +19,125 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+
+	private func cachesDirectory() -> URL {
+		return FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
+	}
+
+	private func trackFromMemoryLeaks(_ instance: AnyObject, file: StaticString = #filePath, line: UInt = #line) {
+		addTeardownBlock { [weak instance] in
+			XCTAssertNil(instance, "Instance should have been deallocated. Potential memory leak.", file: file, line: line)
+		}
+	}
```
caiozullo, line 140: This helper is duplicated in both test classes. You can extract it to another file and reuse it.

caiozullo (commented):
Hi @jorgifumi, 

- All tests are passing ✅ 

Great job!

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +28,76 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+
+		expect(storeToLoad, toRetrieve: .empty)
 	}
 	
 	// - MARK: Helpers
-	
-	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
+
+	private func makeSUT(storeURL: URL? = nil, file: StaticString = #filePath, line: UInt = #line) -> FeedStore {
```
caiozullo, line 75: The `storeURL` parameter is never passed to this function - you can remove it.

jorgifumi: Done! Thank you!

caiozullo:
Hi @jorgifumi - the tests are failing since the last changes. Have a look.

PR #72, opened 2021-01-21
Realm Implementation - Carlos

carlosdamiao:


FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,59 @@
...
+import Foundation
+import RealmSwift
+
+public class RealmFeedStore: FeedStore {
+
+	private let realm: Realm
+
+	public init(config: Realm.Configuration) {
+		realm = try! Realm(configuration: config)
+	}
```
caiozullo, line 18: `try!` can lead to runtime errors - ideally, this initializer should throw errors so clients can handle them.

```suggestion
	public init(config: Realm.Configuration) throws {
		realm = try Realm(configuration: config)
	}
```

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,59 @@
...
+			return
+		}
+
+		switch feed.feedImages.isEmpty {
+		case true:
+			completion(.empty)
+		case false:
+			completion(.found(feed: feed.feedImages.map { $0.asDomain() },
+							  timestamp: feed.timestamp))
+		}
```
caiozullo, line 57: An empty array of images stored in the DB is a valid cache. So the store shouldn't check for `isEmpty`. 

If there's a rule about empty arrays, this check should be in the use case - not in the infrastructure store.

```suggestion
		completion(.found(feed: feed.feedImages.map { $0.asDomain() },
							  timestamp: feed.timestamp))
```

RealmModule/LocalFeedImageRealm.swift
```diff
@@ -0,0 +1,54 @@
...
+		"id"
+	}
+}
+
+public extension LocalFeedImageRealm {
+	func asDomain() -> LocalFeedImage {
+		guard let uuid = UUID(uuidString: id),
+			  let url = URL(string: url) else {
+			fatalError("Data issue when mapping from Realm object to Domain object (LocalFeedImage)")
+		}
```
caiozullo, line 39: `fatalError` would crash the app - ideally, it should return `nil` in case of faulty data.

```suggestion
	func asDomain() -> LocalFeedImage? {
		guard let uuid = UUID(uuidString: id),
			  let url = URL(string: url) else {
			return nil
		}
```

RealmModule/LocalFeedImageRealm.swift
```diff
@@ -0,0 +1,54 @@
...
+//  Created by Carlos Damiao on 16/01/2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+public class LocalFeedImageRealm: Object {
+	@objc dynamic var id = ""
+	@objc dynamic var _description = ""
```
caiozullo, line 14: `description` should be optional:

```suggestion
	@objc dynamic var _description: String? = nil
```

RealmModule/LocalFeedImageRealm.swift
```diff
@@ -0,0 +1,54 @@
...
+	@objc dynamic var location = ""
```
caiozullo, line 15: `location` should be optional:

```suggestion
	@objc dynamic var location: String? = nil
```

RealmModule/LocalFeedImageRealm.swift
```diff
@@ -0,0 +1,54 @@
...
+	func asDomain() -> LocalFeedImage {
+		guard let uuid = UUID(uuidString: id),
+			  let url = URL(string: url) else {
+			fatalError("Data issue when mapping from Realm object to Domain object (LocalFeedImage)")
+		}
+		return LocalFeedImage(id: uuid,
+							  description: !_description.isEmpty ? _description : nil,
+							  location: !location.isEmpty ? location : nil,
+							  url: url)
+	}
```
caiozullo, line 44: The store shouldn't check the strings for `isEmpty`. These properties should be optional as per the requirements (see my other comments).

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +28,67 @@ class FeedStoreIntegrationTests: XCTestCase {
...
-		
+		let sut = makeSUT()
+		sut.deleteCachedFeed { _ in }
 	}
 	
 	private func undoStoreSideEffects() {
-		
+		let sut = makeSUT()
+		sut.deleteCachedFeed { _ in }
+	}
```
caiozullo, line 89: The logic is repeated in both methods. You can create another method to avoid duplication.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +28,67 @@ class FeedStoreIntegrationTests: XCTestCase {
...
-		fatalError("Must be implemented")
+		let config = Realm.Configuration(fileURL: realmURL())
+		let sut = RealmFeedStore(config: config)
+		return sut
 	}
 	
 	private func setupEmptyStoreState() {
-		
+		let sut = makeSUT()
+		sut.deleteCachedFeed { _ in }
```
caiozullo, line 83: Since this operation has an async completion closure, it's important to wait for the completion to be called.

You can use the given helper method that waits for the operation to finish:

```suggestion
		deleteCache(from: sut)
```

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +28,67 @@ class FeedStoreIntegrationTests: XCTestCase {
...
 	}
 	
 	private func undoStoreSideEffects() {
-		
+		let sut = makeSUT()
+		sut.deleteCachedFeed { _ in }
```
caiozullo, line 88: Since this operation has an async completion closure, it's important to wait for the completion to be called.

You can use the given helper method that waits for the operation to finish:

```suggestion
		deleteCache(from: sut)
```

caiozullo (commented):
Hi @carlosdamiao,

Please, check my review comments and let me know when you're done so I can review it again. Thank you 🙂

PR #73, opened 2021-01-22
Alex Nikolaychuk - Feed Store Challenge

alexnikol:


FeedStoreChallenge/FeedStoreImplementations/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,127 @@
...
+//
+
+import Foundation
+import CoreData
+
+public class CoreDataFeedStore: FeedStore {
+	
+	private let container: NSPersistentContainer
+	private let context: NSManagedObjectContext
+	private let modelName = "CoreDataFeedModel"
```
caiozullo, line 16: This property is only used in the initializer. So it shouldn't be an instance property because it doesn't need to exist in memory throughout the lifetime of the instance. It could be a static property.

FeedStoreChallenge/FeedStoreImplementations/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,127 @@
...
+	private let storeURL: URL!
```
caiozullo, line 17: The `storeURL` is only used in the initializer - so it shouldn't be a property.

FeedStoreChallenge/FeedStoreImplementations/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,127 @@
...
+		let description = NSPersistentStoreDescription(url: storeURL)
+		container = NSPersistentContainer(name: modelName, managedObjectModel: model)
+		container.persistentStoreDescriptions = [description]
+				
+		var loadError: Swift.Error?
+		container.loadPersistentStores { loadError = $1 }
+		try loadError.map { throw $0 }
+		
+		context = container.newBackgroundContext()
+	}
```
caiozullo, line 37: This initializer is quite long. You could use the helper methods to make it more concise (see my other comment).

Tests/FeedStoresTests/CoreDataFeedStoreTests/CoreDataFeedStoreIntegrationTests.swift
```diff
@@ -0,0 +1,94 @@
...
+	
+	private func makeSUT() -> FeedStore {
+		let storeURL = testSpecificURL()
+		let sut = try! CoreDataFeedStore(storeURL: storeURL)
+		trackMemoryLeaks(sut)
+		return sut
+	}
+	
+	private func testSpecificURL() -> URL {
+		return URL(fileURLWithPath: "/dev/null2")
```
caiozullo, line 79: In the integration tests, it's important to use a real URL. For example, you can use a file URL within the `cachesDirectory` as shown in the lectures.

FeedStoreChallenge/FeedStoreImplementations/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,127 @@
...
+				} catch {
+					completion(error)
+				}
+
+		}
+	}
+			
+}
+
+private extension NSPersistentContainer {
```
caiozullo, line 99: These helper extension methods are not used. You could use them to make the initializer more concise.

FeedStoreChallenge/FeedStoreImplementations/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,127 @@
...
+		}
+	}
+	
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		let context = self.context
+		context.perform {
+			do {
+				let request = NSFetchRequest<CDCache>(entityName: CDCache.entity().name!)
+				request.returnsObjectsAsFaults = false
+				if let foundCache = try context.fetch(request).first {
```
caiozullo, line 62: This exact same logic exists in two places (in the `retrieve` and `deleteCachedFeed` methods). 

You can extract it into a static method on `CDCache` to eliminate duplication. Something like:

```
if let foundCache = try CDCache.findFirst(in: context) {
  ...
}
```

FeedStoreChallenge/FeedStoreImplementations/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,127 @@
...
+		let context = self.context
+			self.deleteCachedFeed { error in
+				guard error == nil else { return completion(error) }
+				do {
+					let cache = CDCache(context: context)
+					cache.timeStamp = timestamp
+					cache.feed = NSOrderedSet(array: feed.map { local in
+						let feed = CDFeedImage(context: context)
+						feed.from(local)
+						return feed
```
caiozullo, line 85: `from` doesn't read like an action - it reads like it will return something, but it doesn't. 

It'd be best to create a `convenience` initializer, so the `CDFeedImage` is instantiated with the correct values. Something like:

```
return CDFeedImage(local: local, context: context)
```

caiozullo (commented):
Hi @alexnikol,

- All tests are passing ✅ 

Great job!

PR #74, opened 2021-01-24
Wilmer Barrios - Feed Store Challenge

wilbarrios:


FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,121 @@
...
+import Foundation
+import CoreData
+
+public final class CoreDataFeedStore: FeedStore {
+	
+	typealias StoredFeedData = (feed: [CDFeedImageItem], timestamp: Date)
+	
+	private let container: NSPersistentContainer
+	private let context: NSManagedObjectContext
+	private let modelName = "FeedStoreDataModel"
```
caiozullo, line 18: This property is only used in the initializer. So it shouldn't be a property because it doesn't need to exist in memory throughout the lifetime of the instance.

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,121 @@
...
+	private let storeURL: URL!
```
caiozullo, line 19: The `storeURL` is only used in the initializer - so it shouldn't be a property.

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,121 @@
...
+		let description = NSPersistentStoreDescription(url: storeURL)
+		container = NSPersistentContainer(name: modelName, managedObjectModel: model)
+		container.persistentStoreDescriptions = [description]
+		
+		var loadError: Swift.Error?
+		container.loadPersistentStores { loadError = $1 }
+		try loadError.map { throw $0 }
+		
+		context = container.newBackgroundContext()
+	}
```
caiozullo, line 39: This initializer is quite long. You could create some helper methods to make it more concise.

FeedStoreChallenge/Core Data/Entities/CDFeedImage+CoreDataProperties.swift
```diff
@@ -0,0 +1,62 @@
...
+
+
+extension CDFeedImage {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<CDFeedImage> {
+        return NSFetchRequest<CDFeedImage>(entityName: "CDFeedImage")
+    }
+
+    @NSManaged public var feed_timestamp: Date?
+	@NSManaged public var feed_data: NSOrderedSet?
```
caiozullo, line 21: Those properties should *not* be optional.

FeedStoreChallenge/Core Data/Entities/CDFeedImageItem+CoreDataProperties.swift
```diff
@@ -0,0 +1,29 @@
...
+import CoreData
+
+
+extension CDFeedImageItem {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<CDFeedImageItem> {
+        return NSFetchRequest<CDFeedImageItem>(entityName: "CDFeedImageItem")
+    }
+
+    @NSManaged public var image_id: UUID?
```
caiozullo, line 20: The `id` shouldn't be optional.

FeedStoreChallenge/Core Data/Entities/CDFeedImageItem+CoreDataProperties.swift
```diff
@@ -0,0 +1,29 @@
...
+    @NSManaged public var image_description: String?
+    @NSManaged public var image_location: String?
+    @NSManaged public var image_url: URL?
```
caiozullo, line 23: The `url` shouldn't be optional.

FeedStoreChallenge/Core Data/FeedStoreDataModel.xcdatamodeld/FeedStoreDataModel.xcdatamodel/contents
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="17511" systemVersion="19H15" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="CDFeedImage" representedClassName=".CDFeedImage" syncable="YES">
+        <attribute name="feed_timestamp" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
```
caiozullo, line 4: The `feed_timestamp` shouldn't be optional.

FeedStoreChallenge/Core Data/FeedStoreDataModel.xcdatamodeld/FeedStoreDataModel.xcdatamodel/contents
```diff
@@ -0,0 +1,17 @@
...
+        <relationship name="feed_data" optional="YES" toMany="YES" deletionRule="Nullify" ordered="YES" destinationEntity="CDFeedImageItem"/>
```
caiozullo, line 5: The `feed_data` relationship `deletionRule` should be `Cascade` to prevent orphaned objects in the DB.

FeedStoreChallenge/Core Data/FeedStoreDataModel.xcdatamodeld/FeedStoreDataModel.xcdatamodel/contents
```diff
@@ -0,0 +1,17 @@
...
+    </entity>
+    <entity name="CDFeedImageItem" representedClassName=".CDFeedImageItem" syncable="YES">
+        <attribute name="image_description" optional="YES" attributeType="String"/>
+        <attribute name="image_id" optional="YES" attributeType="UUID" usesScalarValueType="NO"/>
```
caiozullo, line 9: The `image_id` shouldn't be optional.

FeedStoreChallenge/Core Data/FeedStoreDataModel.xcdatamodeld/FeedStoreDataModel.xcdatamodel/contents
```diff
@@ -0,0 +1,17 @@
...
+        <attribute name="image_location" optional="YES" attributeType="String"/>
+        <attribute name="image_url" optional="YES" attributeType="URI"/>
```
caiozullo, line 11: The `image_url` shouldn't be optional.

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,121 @@
...
+			} catch let e {
+				completion(e)
+			}
+		}
+	}
+	
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		let context = self.context
+		context.perform {
+			let fetchRequest: NSFetchRequest<CDFeedImage> = CDFeedImage.fetchRequest()
```
caiozullo, line 74: I believe the compiler can infer the type:

```suggestion
			let fetchRequest = CDFeedImage.fetchRequest()
```

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,121 @@
...
+		container.loadPersistentStores { loadError = $1 }
+		try loadError.map { throw $0 }
+		
+		context = container.newBackgroundContext()
+	}
+	
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		let context = self.context
+		context.perform {
+			let fetchRequest: NSFetchRequest<NSFetchRequestResult> = CDFeedImage.fetchRequest()
```
caiozullo, line 44: I believe the compiler can infer the type:

```suggestion
			let fetchRequest = CDFeedImage.fetchRequest()
```

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,121 @@
...
+			}
+		}
+	}
+	
+	// MARK: Helpers
+	private func map(_ data: [CDFeedImage]) -> (items: [LocalFeedImage], timestamp: Date) {
+		var result = [StoredFeedData]()
+		for feed in data {
+			guard let feedData = feed.feed_data, let feedTimestamp = feed.feed_timestamp else { continue }
+			result.append((feedData.array as! [CDFeedImageItem], feedTimestamp))
```
caiozullo, line 92: Force-casting with `as!` can lead to runtime errors. Instead, it could be part of the guard:

```suggestion
			guard let feedData = feed.feed_data.array as? [CDFeedImageItem], 
			  let feedTimestamp = feed.feed_timestamp else { continue }
			result.append((feedData, feedTimestamp))
```

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,121 @@
...
+		let mappedData = result.last!.feed.toLocal()
+		let fetchedTimestamp = result.last!.timestamp
+		return (mappedData, fetchedTimestamp)
+	}
+}
+
+// MARK: Mapping extensions
+internal extension Array where Element == CDFeedImageItem {
+	func toLocal() -> [LocalFeedImage] {
+		filter({ $0.image_id != nil && $0.image_url != nil }).map({ LocalFeedImage(id: $0.image_id!, description: $0.image_description, location: $0.image_location, url: $0.image_url! ) })
```
caiozullo, line 104: Instead of `filter` then `map`, you can use `compactMap` to simplify the logic.

But in this case, the `image_id` and `image_url` shouldn't be optional (see my other comments) so the filter is unnecessary.

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,121 @@
...
+	}
+	
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		let context = self.context
+		context.perform {
+			let fetchRequest: NSFetchRequest<CDFeedImage> = CDFeedImage.fetchRequest()
+			do {
+				let data = try context.fetch(fetchRequest)
+				let mappedData = self.map(data)
+				guard !mappedData.items.isEmpty else { completion(.empty); return }
```
caiozullo, line 78: At the moment, the `map` method is too complex. It does too many things.

The guard should be **before** calling `map` - so the implementation can be much simpler.

```suggestion
				guard let feed = data.first else { return completion(.empty) }
				let mappedData = self.map(feed)
```

Moreover, the `map` method doesn't require any instance method/property. So it could be a `static func`. This way, you don't need to capture `self` within the closure.

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,121 @@
...
+				completion(.found(feed: mappedData.items, timestamp: mappedData.timestamp))
+			} catch let error as NSError {
+				completion(.failure(error))
+				print("Could not fetch. \(error), \(error.userInfo)")
```
caiozullo, line 82: Remote the unnecessary print statement.

caiozullo (commented):
Hi @wilbarrios,

Please, check my review comments and let me know when you're done so I can review it again. Thank you 🙂

CDFeedImage+CoreDataClass.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  CDFeedImage+CoreDataClass.swift
```
caiozullo, line 2: There are duplicate and unused files, including this one. You can remove them.

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
wilbarrios: Hi Caio! Thanks for all your feed back.

This particular line give me compiler error, evidence attached.

image

FeedStoreChallenge/Core Data/FeedStoreDataModel.xcdatamodeld/FeedStoreDataModel.xcdatamodel/contents
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="17511" systemVersion="19H15" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="CDFeedImage" representedClassName=".CDFeedImage" syncable="YES">
+        <attribute name="feed_timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feed_data" toMany="YES" deletionRule="Nullify" ordered="YES" destinationEntity="CDFeedImageItem"/>
```
caiozullo, line 5: The `feed_data` relationship `deletionRule` should be `Cascade` to prevent orphaned objects in the DB.

FeedStoreChallenge/Core Data/FeedStoreDataModel.xcdatamodeld/FeedStoreDataModel.xcdatamodel/contents
```diff
@@ -0,0 +1,17 @@
...
+    </entity>
+    <entity name="CDFeedImageItem" representedClassName=".CDFeedImageItem" syncable="YES">
+        <attribute name="image_description" optional="YES" attributeType="String"/>
+        <attribute name="image_id" attributeType="UUID" usesScalarValueType="NO"/>
+        <attribute name="image_location" optional="YES" attributeType="String"/>
+        <attribute name="image_url" attributeType="URI"/>
+    </entity>
```
caiozullo, line 12: The `CDFeedImageItem` entity is missing the inverse relationship with the `CDFeedImage` enity.

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,101 @@
...
+import Foundation
+import CoreData
+
+public final class CoreDataFeedStore: FeedStore {
+	
+	typealias StoredFeedData = (feed: [CDFeedImageItem], timestamp: Date)
+	
+	private let context: NSManagedObjectContext
+	
+	public init(contextBuilder: CDContextBuilder) throws {
```
caiozullo, line 18: This initializer should receive the `storeURL`. It could then create the `CDContextBuilder` and do the logic. This way, you don't move this responsibility to clients because that would be a private implementation detail.

FeedStoreChallenge/Core Data/CoreDataInstanceBuilder.swift
```diff
@@ -0,0 +1,69 @@
...
+
+extension CoreDataInstanceBuilder: CDContextBuilder {
+	public func build() -> CDContextBuildResult {
+		do {
+			let context = try makeCDContext(storeURL: self.storeURL)
+			return .success(context)
+		} catch let e {
+			return .failure(e)
+		}
+	}
```
caiozullo, line 68: Since this function is synchronous, it should throw errors in case of failures, instead of returning errors:

See how much simpler it'd be:

```suggestion
	public func build() throws -> NSManagedObjectContext {
		return try makeCDContext(storeURL: self.storeURL)
	}
```

FeedStoreChallenge/Core Data/CoreDataInstanceBuilder.swift
```diff
@@ -0,0 +1,69 @@
...
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+public enum CDContextBuildResult {
+	case success(NSManagedObjectContext)
+	case failure(Error)
+}
```
caiozullo, line 15: This type is unnecessary. See my other comment.

FeedStoreChallenge/Core Data/CoreDataInstanceBuilder.swift
```diff
@@ -0,0 +1,69 @@
...
+
+public protocol CDContextBuilder {
+	func build() -> CDContextBuildResult
+}
```
caiozullo, line 19: You don't need a protocol if this class is not used by clients. See my other comments.

FeedStoreChallenge/Core Data/CoreDataInstanceBuilder.swift
```diff
@@ -0,0 +1,69 @@
...
+
+public final class CoreDataInstanceBuilder {
```
caiozullo, line 21: This class doesn't need to be public if it's not used by clients. It can be internal. See my other comments.

FeedStoreChallenge/Core Data/CoreDataInstanceBuilder.swift
```diff
@@ -0,0 +1,69 @@
...
+	
+	private static let modelName = "FeedStoreDataModel"
+	let storeURL: URL
```
caiozullo, line 24: This property can be private.

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,101 @@
...
+			} catch let e {
+				completion(e)
+			}
+		}
+	}
+	
+	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+		let context = self.context
+		context.perform {
+			let feed_data: [CDFeedImageItem] = feed.toCD(context: context)
```
caiozullo, line 46: The compiler can infer the type:

```suggestion
			let feed_data = feed.toCD(context: context)
```

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,101 @@
...
+				completion(.failure(error))
+			}
+		}
+	}
+	
+	// MARK: Helpers
+	private static func map(_ feed: CDFeedImage) -> (items: [LocalFeedImage], timestamp: Date) {
+		guard let feedData = feed.feed_data.array as? [CDFeedImageItem] else { return ([], Date()) }
+		return (feedData.toLocal(), feed.feed_timestamp)
+	}
```
caiozullo, line 78: The store is an infrastructure detail and it shouldn't contain any logic. Logic should be somewhere else.

The store implementation should just translate the values to and from the DB.

So it should never replace values with default values like `[]` and `Date()` because it can lead to unexpected behavior. Moreover, the `map` function is unnecessary this logic can be done inline (see my other comment).

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,101 @@
...
+	
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		let context = self.context
+		context.perform {
+			let fetchRequest: NSFetchRequest<CDFeedImage> = CDFeedImage.fetchRequest()
+			do {
+				let data = try context.fetch(fetchRequest)
+				guard let feed = data.last else { return completion(.empty) }
+				let (items, timestamp) = CoreDataFeedStore.map(feed)
+				completion(.found(feed: items, timestamp: timestamp))
```
caiozullo, line 67: You can combine the guard checks into one, so you won't need the `map` function:

```suggestion
				let data = try context.fetch(fetchRequest)
				guard let feed = data.last, let items = feed.feed_data.array as? [CDFeedImageItem] else { return completion(.empty) }
				
				completion(.found(feed: items.toLocal(), timestamp: feed.feed_timestamp))
```

caiozullo (commented):
Hi @wilbarrios, almost there!

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
wilbarrios: Thanks @caiozullo for all this feedback!

I have a question for this point, I've read about better compile times (the app I'm working on, has 450 - 500 seconds on clean build) and one suggestion was to explicitly write properties/variables types. Is this suggestion very beneficial? or doesn't worth it? What would be your suggestion?. Making MacOS modules seems really fast for large applications with high compile times. Thanks again!

https://medium.com/rocket-fuel/optimizing-build-times-in-swift-4-dc493b1cc5f5

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,89 @@
...
+//  Created by Wilmer Barrios on 23/01/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+public final class CoreDataFeedStore: FeedStore {
+	
+	typealias StoredFeedData = (feed: [CDFeedImageItem], timestamp: Date)
```
caiozullo, line 14: This `typealias` is not used - you can remove it.

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
caiozullo: It's beneficial to add the types if the compiler takes a long time to infer it (e.g., complex syntax and chained closures are usually slow). My suggestion is to measure before optimizing - you can use build settings to find the slow places. Moreover, new Swift versions are getting better at it ✅

caiozullo (commented):
Hi @wilbarrios,

- All tests are passing ✅ 

I left one final review comment. Feel free to push changes.

Great job!

FeedStoreChallenge/Core Data/CoreDataFeedStore.swift
wilbarrios: Thanks for your feedback!

I'll be checking build times measure methods, have a good day!

PR #77, opened 2021-01-28
Oliver Perez - Feed Store Challenge

oliver-perez:
This is my solution for The FeedStore Challenge using Core Data.


Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,137 +20,101 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+			super.init(modelName: modelName, bundle: bundle)
+
+			let persistentStoreDescription = NSPersistentStoreDescription()
+			persistentStoreDescription.type = NSInMemoryStoreType
+
+			let container = NSPersistentContainer.loadModel(name: modelName, in: bundle, descriptions: [persistentStoreDescription])
+
+			self.storeContainer = container
+		}
+	}
```
caiozullo, line 111: A simpler way to create an in-memory CoreData store is to use a file URL to `/dev/null` as shown in the lectures.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,137 @@
...
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+public final class CoreDataFeedStore: FeedStore {
+	
+	// MARK: Properties
+	lazy var context: NSManagedObjectContext = {
```
caiozullo, line 15: This property can be private - it's only used within this class.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,137 @@
...
+		Self.coreDataStack.context
+	}()
+	private static var coreDataStack: CoreDataStack = CoreDataStack(modelName: "FeedStoreModel")
```
caiozullo, line 18: This property can be a `let` - it never changes. 

Moreover, it shouldn't be `static`. Otherwise, every time a new `CoreDataFeedStore` instance is created, the whole stack is replaced for all instances - and that can lead to unexpected behavior!

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,137 @@
...
+	
+	// MARK: Initializers
+	public init(coreDataStack: CoreDataStack) {
+		Self.coreDataStack = coreDataStack
+	}
```
caiozullo, line 23: This initializer could receive the `storeURL`. It could then create the `CoreDataStack` as an internal/private detail. This way, clients don't need to deal with `CoreDataStack` - it's a private detail. So you can also refactor it without breaking clients or tests! ✅

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,137 @@
...
+	
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		context.perform { [unowned context] in
+			
+			do {
+				guard let cache = try self.fetchCache() else { return completion(nil) }
+				context.delete(cache)
+				
+				completion(nil)
```
caiozullo, line 32: It should only complete after saving the operation:

```suggestion
				context.delete(cache)
				try context.save()
				completion(nil)
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,137 @@
...
+				
+			} catch {
+				return completion(error)
+			}
+			
+			Self.coreDataStack.saveContext()
```
caiozullo, line 38: This line should be removed here - it should save the context before calling the completion closure because the `save` operation can fail.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,137 @@
...
+			
+			feedCache.feedImages = NSOrderedSet(array: feedImages)
+			feedCache.timestamp = timestamp
+			
+			do {
+				try self.context.save()
+			} catch {
+				completion(error)
+			}
+			completion(nil)
```
caiozullo, line 65: The `completion` closure should only be called once, with either an `error` or `nil`. But this method can call the `completion` closure more than once.

To solve it and simplify the logic, the insert operation should all be within the do/catch block, since the `save` operation can fail:

```suggestion
			do {
				if let cache = try self.fetchCache() {
					self.context.delete(cache)
				}
				
				let feedCache = FeedCache(context: self.context)
				let feedImages = transformToManagedRepresentation(feed)
				
				feedCache.feedImages = NSOrderedSet(array: feedImages)
				feedCache.timestamp = timestamp
				try self.context.save()
				completion(nil)
			} catch {
				completion(error)
			}
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,137 @@
...
+			
+			let localFeedResult = cache.map {
+				(feed: $0.feedImages, timestamp: $0.timestamp)
+			}
+			
+			guard let feedResult = localFeedResult,
+						let timestamp = localFeedResult?.timestamp
+			else { return completion(.empty) }
+			
+			let feed = feedResult.feed?.array as? [FeedImage] ?? []
```
caiozullo, line 82: This logic can be simplified in one line:

```suggestion
			guard let cache = try fetchCache(), let feed = cache.feedImages?.array as? [FeedImage] else { return completion(.empty) }
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,137 @@
...
+			feedImage.imageDescription = $0.description
+			feedImage.url = $0.url
+			
+			return feedImage
+		}
+	}
+	
+	private func fetchCache() throws -> FeedCache? {
+		let fetchRequest: NSFetchRequest<FeedCache> = FeedCache.fetchRequest()
+		let result = try Self.coreDataStack.context.fetch(fetchRequest)
```
caiozullo, line 112: Should use `self.context`, since all other operations are using the instance context.

```suggestion
		let result = try self.context.fetch(fetchRequest)
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,137 @@
...
+	
+}
+
+public extension NSPersistentContainer {
+	
+	// MARK: NSPersistentContainer extension
+	
+	static func loadModel(name: String, in bundle: Bundle, descriptions: [NSPersistentStoreDescription] = [NSPersistentStoreDescription()]) -> NSPersistentContainer {
+		guard let mom = NSManagedObjectModel.mergedModel(from: [bundle]) else {
+			fatalError("Not able to find data model")
```
caiozullo, line 125: Errors should be handled. Don't use `fatalError` as it can lead to runtime errors.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,137 @@
...
+		}
+		
+		let container = NSPersistentContainer(name: name, managedObjectModel: mom)
+		container.persistentStoreDescriptions = descriptions
+		container.loadPersistentStores { d, e in
+			_ = e.flatMap {_ in
+				fatalError("Not able to load store")
```
caiozullo, line 132: Errors should be handled. Don't use `fatalError` as it can lead to runtime errors.

FeedStoreChallenge/CoreDataManagedObjects/FeedCache+CoreDataProperties.swift
```diff
@@ -0,0 +1,62 @@
...
+import CoreData
+
+
+extension FeedCache {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<FeedCache> {
+        return NSFetchRequest<FeedCache>(entityName: "FeedCache")
+    }
+
+    @NSManaged public var timestamp: Date?
```
caiozullo, line 20: `timestamp` shouldn't be optional.

FeedStoreChallenge/CoreDataManagedObjects/FeedCache+CoreDataProperties.swift
```diff
@@ -0,0 +1,62 @@
...
+    @NSManaged public var feedImages: NSOrderedSet?
```
caiozullo, line 21: `feedImages` shouldn't be optional.

FeedStoreChallenge/CoreDataManagedObjects/FeedImage+CoreDataProperties.swift
```diff
@@ -0,0 +1,30 @@
...
+import CoreData
+
+
+extension FeedImage {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<FeedImage> {
+        return NSFetchRequest<FeedImage>(entityName: "FeedImage")
+    }
+
+    @NSManaged public var id: UUID?
```
caiozullo, line 20: `id` shouldn't be optional.

FeedStoreChallenge/CoreDataManagedObjects/FeedImage+CoreDataProperties.swift
```diff
@@ -0,0 +1,30 @@
...
+    @NSManaged public var imageDescription: String?
+    @NSManaged public var location: String?
+    @NSManaged public var url: URL?
```
caiozullo, line 23: `url` shouldn't be optional.

FeedStoreChallenge/CoreDataStack.swift
```diff
@@ -0,0 +1,40 @@
...
+import Foundation
+import CoreData
+
+open class CoreDataStack {
+
+	let modelName: String
+	let bundle: Bundle
+	
+	public lazy var context: NSManagedObjectContext = {
+		storeContainer.viewContext
```
caiozullo, line 18: The `viewContext` is tied to the main thread. So executing CoreData operations here can be expensive. Ideally, you should use a background context to avoid blocking the main thread (e.g., blocking the main thread makes the UI unresponsive.).

FeedStoreChallenge/CoreDataStack.swift
```diff
@@ -0,0 +1,40 @@
...
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+open class CoreDataStack {
+
+	let modelName: String
+	let bundle: Bundle
```
caiozullo, line 15: Those properties should be private - they're only used within this class.

FeedStoreChallenge/CoreDataStack.swift
```diff
@@ -0,0 +1,40 @@
...
+	
+	public lazy var context: NSManagedObjectContext = {
```
caiozullo, line 17: This property doesn't need to be public - it can be internal.

FeedStoreChallenge/CoreDataStack.swift
```diff
@@ -0,0 +1,40 @@
...
+		storeContainer.viewContext
+	}()
+	
+	public lazy var storeContainer: NSPersistentContainer = {
```
caiozullo, line 21: This property doesn't need to be public - it can be internal.

FeedStoreChallenge/CoreDataStack.swift
```diff
@@ -0,0 +1,40 @@
...
+	
+	public func saveContext() {
+		context.perform { [weak self] in
+			guard let self = self else { return }
+			do {
+				try self.context.save()
+			} catch let error as NSError {
+				fatalError("Unresolved error \(error), \(error.userInfo)")
+			}
+		}	}
```
caiozullo, line 38: This method is unnecessary - you can remove it and perform `save` directly in the context.

Moreover, it should never use `fatalError` and it leads to runtime errors.

FeedStoreChallenge/CoreDataStack.swift
```diff
@@ -0,0 +1,40 @@
...
+	
+	public lazy var context: NSManagedObjectContext = {
+		storeContainer.viewContext
+	}()
+	
+	public lazy var storeContainer: NSPersistentContainer = {
+		NSPersistentContainer.loadModel(name: modelName, in: bundle)
+	}()
+
+	public init(modelName: String, bundle: Bundle = .main) {
```
caiozullo, line 25: The `modelName` should be an internal detail. Clients shouldn't need to know the model name to use this class.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +27,68 @@ class FeedStoreIntegrationTests: XCTestCase {
...
 	}
 	
 	private func undoStoreSideEffects() {
-		
+		deleteStoreArtifacts()
+	}
+	
+	private func deleteStoreArtifacts() {
+		let sut = makeSUT()
+		sut.deleteCachedFeed { _ in }
```
caiozullo, line 91: The `deleteCachedFeed` is an async operation. So it's important to wait for the completion. Otherwise, another test might start before the operation finishes - leading to race conditions!

You can use the provided `deleteCache(from: sut)` helper that waits for the operation to finish:

```suggestion
		deleteCache(from: sut)
```

caiozullo (commented):
Hi @oliver-perez,

PR #79, opened 2021-01-31
Bogdan Poplauschi - Feed Store Challenge

bpoplauschi:
# Intent
Implement concrete `FeedStore` types using TDD.

# Implementation

## InMemoryFeedStore
- simple implementation of the `FeedStore` protocol using an in-memory single reference cache
- internally using private `Cache` type to group the data (feed and timestamp)
- this implementation cannot fail nor persist, so it has no failable nor integration tests
- tests file is `InMemoryFeedStoreTests`

## CoreDataFeedStore
- created `CoreDataFeedStore` CoreData implementation of `FeedStore`
- by applying the Single Responsibility Principle, here are the resulted types and functions:
  - internal helper class `CoreDataFeedHelper` maps between the CoreData models `CDFeed` /  `CDFeedImage` and our `Feed Feature` models `[LocalFeedImage]` + `Date`
  - `NSPersistentContainer` and `NSManagedObjectModel` extensions for setting up the store
  - `CDFeed` private extension for creating `NSFetchRequest`
  - `XCTestCase+Helpers` extension for sharing helpers
  - `XCTestCase+MemoryLeaksTracker` extension for sharing memory leak tracking code
- added extra test `test_init_throwsErrorWhenManagedObjectModelCannotBeCreated` to assert we throw the proper error when the managed object model cannot be created
- test files are `CoreDataFeedStoreTests` and `CoreDataFeedStoreIntegrationTests`

# Class Diagram

classDiagram

Note: the diagram was created using [mermaid](https://mermaid-js.github.io/mermaid/) which uses a markdown syntax. The `classDiagram.md` contains the "source code" which is usually easy to maintain if added to the workspace.
It's not as nice as one created using [draw.io](draw.io), but it's easier to maintain IMO.

# Mocking CoreData PersistentStore

I spent some hours trying to mock the CoreData `PersistentStore` by creating an `NSIncrementalStore` subclass, so we can mock fail the `CoreDateFeedStore`.

Unfortunately, this approach is flaky and unreliable. Even if the docs state `NSIncrementalStore` and `NSPersistentStore`are abstract classes, they definitely have some behavior we just can't override.

If you want to take a deeper look, my attempt can be found on [my fork's mock_coredate_persistentstore branch](https://github.com/bpoplauschi/ios-lead-essentials-feed-store-challenge/tree/mock_coredata_persistentstore).
For this PR, decided not to include that attempt which I don't think is deliverable code.


FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,115 @@
...
+	}
+	
+	// MARK: - Properties
+	
+	private let persistentContainer: NSPersistentContainer
+	private let managedContext: NSManagedObjectContext
+	
+	// MARK: - Init
+	
+	public init(storeURL: URL, modelName: String? = nil) throws {
```
caiozullo, line 26: The `modelName` should be a private detail - so clients don't need to know the actual model name. 

This store probably wouldn't work with other object models, so it'd be best to prevent clients from passing the wrong values here.

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+	
+	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+		let context = self.managedContext
+		
+		context.perform {
+			do {
+				try self.deleteCache()
+				let _ = CoreDataFeedMapper.mapToStorableFeed(feed: feed, timestamp: timestamp, in: context)
```
caiozullo, line 57: `map` is usually an operation without side-effects. But since the result here is ignored with `let _`, it's implied that there are side-effects (objects added to the managed context).

Maybe it'd be best to rename `mapToStorableFeed` to `insert` to clarify the side-effects:

```
CoreDataFeedMapper.insert(feed: feed, timestamp: timestamp, in: context)
```

Moreover, you don't need to return anything in this function since it's purely performing side-effects. Thus, you don't need to ignore the result. What do you think?

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+}
+
+private extension NSPersistentContainer {
+	convenience init(dataModelName: String, model: NSManagedObjectModel, storeURL: URL) {
+		let description = NSPersistentStoreDescription(url: storeURL)
+		self.init(name: dataModelName, managedObjectModel: model)
+		persistentStoreDescriptions = [description]
+		loadPersistentStores { _, _ in }
```
caiozullo, line 98: `loadPersistentStores` can fail, but the result is not handled.

This means this class may be instantiated in an invalid state - leading to undefined behavior at runtime.

Ideally, if it fails to `loadPersistentStores`, it should throw an error so clients can handle it.

Tests/CoreDataFeedStoreIntegrationTests.swift
```diff
@@ -0,0 +1,60 @@
...
+
+        deleteCache(from: storeToDelete)
+
+        expect(storeToLoad, toRetrieve: .empty)
+	}
+	
+	// - MARK: Helpers
+	
+	private func makeSUT() -> FeedStore {
+		let sut = try! CoreDataFeedStore(storeURL: URL(fileURLWithPath: "/dev/null").appendingPathComponent("shared"))
```
caiozullo, line 56: In these integration tests, you could use a real URL to test the store persisting data to disk. 

`/dev/null` is fine for isolated tests. But in my experience, the in-memory store doesn't always behave exactly like the real thing - so bugs may pass unnoticed in integration tests.

Tests/CoreDataFeedStoreTests.swift
```diff
@@ -0,0 +1,103 @@
...
+		} catch {
+			XCTAssertEqual(
+				error as NSError,
+				CoreDataFeedStore.CoreDataFeedStoreError.cannotCreateManagedObjectModel(
+					name: "WrongModelName",
+					bundle: Bundle(for: CoreDataFeedStore.self)
+				) as NSError
+			)
+		}
+	}
```
caiozullo, line 22: This test is not needed if the `modelName` is a private detail (see my other comment).

caiozullo (commented):
Hi @bpoplauschi,

- All tests are passing ✅ 
- Bonus points for the Integration Tests, Memory Leak Tracking, and Diagrams ✅ 

Great job!

Regarding your comments, I also found `NSIncrementalStore` to be unstable. There are other ways to simulate errors, like using managed context subclasses or using read-only SQLite stores (so you can't insert/delete, for example). But they are usually too expensive to create. I would use them only if needed, e.g., if there's a regression related to such errors.

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
bpoplauschi: I understand and agree, I didn't follow CQS.
My reasoning was the `CoreDataFeedMapper` would just handle mapping between `CDFeed` / `CDFeedImage` and `LocalFeedImage` + timestamp.

Since `insert` is now a command (renamed and removed return type), I decided to also rename `CoreDataFeedMapper` to `CoreDataFeedHelper`.

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
bpoplauschi: I've redone this part now, after going through the last lecture of the Persistence chapter (very nice).
Decided to replace the `NSPersistentContainer` nullable initializer with a static func `load` that throws, that way we can better handle the error cases (didn't want to have an init that can return nil and throw, that is just confusing).

Tests/CoreDataFeedStoreIntegrationTests.swift
bpoplauschi: I got through the last lecture of the Persistence module and I now understand what you refer to as Integration Tests and the fact that they should not use test doubles - I fully agree.
Using an in memory store is driving a different behavior, so it should not be done in an integration test. Changed it back to using a real test specific storeURL.

bpoplauschi:
Hi @caiozullo,

Thanks for thoroughly reviewing this PR. I wanted to, but could not object to any of your comments :) so I fixed them all

> Regarding your comments, I also found NSIncrementalStore to be unstable. There are other ways to simulate errors, like using managed context subclasses or using read-only SQLite stores (so you can't insert/delete, for example). But they are usually too expensive to create. I would use them only if needed, e.g., if there's a regression related to such errors.

Thanks for responding to this, I was a bit unsatisfied by not being able to test the errors with an `NSIncrementalStore` mock.
I also found mentions about subclassing managed contexts, didn't even think about using read-only SQLite stores.

I agree with your assessment: those are expensive to create (and probably maintain) and the benefits are limited, so unless one has real issues in their projects with this, they're probably not worth doing.

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
caiozullo: That's it. And if the private implementation uses the wrong name, the test will fail. ✅

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
caiozullo: Sounds good. 

I usually create those helpers as extension methods on the managed objects subclasses. For example, `CDFeed.insert(...)`. Or even as convenience initializers `CDFeed(items: items: timestamp: timestamp, context: context)`.

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,92 @@
...
+
+import CoreData
+
+public final class CoreDataFeedStore: FeedStore {
+	
+	// MARK: - Types
+	
+	public enum CoreDataFeedStoreError: Error {
+		case cannotCreateManagedObjectModel(name: String, bundle: Bundle)
+	}
```
caiozullo, line 17: Looks like this type is not used - you can remove it.

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
bogdan-poplauschi-coda: Cool, that is a more elegant approach than the helper I created (after changing it from a more focused mapper). But since it's only an exercise, I'll leave it as is to reflect my original idea. As we would say in a real project: we can always refactor it later :)

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
bogdan-poplauschi-coda: Nice catch

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
bpoplauschi: Cool, that is a more elegant approach than the helper I created (after changing it from a more focused mapper). But since it's only an exercise, I'll leave it as is to reflect my original idea. As we would say in a real project: we can always refactor it later :)

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
bpoplauschi: Nice catch - fixed

caiozullo:
Nice one!

PR #81, opened 2021-02-09
Realm Implementation - Engelman Marko

engelmanmarko:


FeedStoreChallenge/Infrastructure/RealmFeedImage.swift
```diff
@@ -0,0 +1,29 @@
...
+import Foundation
+import RealmSwift
+
+class RealmFeedImage: Object {
+	@objc dynamic var id = ""
+	@objc dynamic var imageDescription: String? = nil
+	@objc dynamic var location: String? = nil
+	@objc dynamic var url = ""
+
+	convenience init(id: String, description: String?, location: String?, url: URL) {
```
caiozullo, line 18: `id` should be a `UUID` to prevent clients from passing *any* `String`

FeedStoreChallenge/Infrastructure/RealmFeedStore.swift
```diff
@@ -0,0 +1,60 @@
...
+//
+//  Created by Marko Engelman on 09/02/2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+public final class RealmFeedStore: FeedStore {
+	let realm: Realm
```
caiozullo, line 13: This property should be private.

FeedStoreChallenge/Infrastructure/RealmFeedStore.swift
```diff
@@ -0,0 +1,60 @@
...
+	
+	enum RealmFeedStore: Error {
+		case unableToInitializeRealmDB
+	}
```
caiozullo, line 17: This error type is not needed - this class can throw the `Realm` errors directly.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +28,70 @@ class FeedStoreIntegrationTests: XCTestCase {
...
 	}
 	
 	private func setupEmptyStoreState() {
-		
+		try? FileManager.default.removeItem(at: testsRealmStoreURL())
 	}
 	
 	private func undoStoreSideEffects() {
 		
 	}
```
caiozullo, line 86: Missing implementation - should undo side-effects

caiozullo (commented):
Hi @engelmanmarko,

- All tests are passing ✅ 

Great job!

FeedStoreChallenge/Infrastructure/RealmFeedImage.swift
engelmanmarko: Thank you for comment :) I wasn't sure about this decision from beginning :) I have updated the code.

FeedStoreChallenge/Infrastructure/RealmFeedStore.swift
engelmanmarko: OK, updated :)

FeedStoreChallenge/Infrastructure/RealmFeedStore.swift
```diff
@@ -0,0 +1,56 @@
...
+
+public final class RealmFeedStore: FeedStore {
+	private let realm: Realm
+	
+	public init(configuration: Realm.Configuration) throws {
+		do {
+			realm = try Realm(configuration: configuration)
+		} catch let error {
+			throw error
+		}
```
caiozullo, line 20: There's no need for the do/catch block here since this method can throw.

```suggestion
		realm = try Realm(configuration: configuration)
```

FeedStoreChallenge/Infrastructure/LocalFeedImageMapper.swift
```diff
@@ -0,0 +1,44 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Marko Engelman on 09/02/2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+
+internal final class LocalFeedImageMapper {
+	internal static func map(feed: [LocalFeedImage], timestamp: Date) -> RealmFeedItems {
```
caiozullo, line 12: As a final comment: Instead of static methods of a separate mapper class, I usually define those methods directly on the Realm models (they can be extensions). For example, a convenience initializer:

`RealmFeedItems(feed: [LocalFeedImage], timestamp: Date)`

engelmanmarko: I have seen this in course as well, but I would go with mapper if you agree. I really like this similarity as we have it when we are mapping remote objects.

caiozullo:
Nice one!

PR #82, opened 2021-02-10
InMemory implementation - Hugo Alonso Luis

halonsoluis:
Hi Caio.

Here is presented a case of an underlying static data structure that will hold the cache memory during the application lifetime.

As to avoid retention of the handler class I have decided to go for reducing the amount of static content and concentrate it only on the shared data structure. In this case an array of CacheData that is composed by the feed and the timestamp.

The main objective on this was to mimic as much as possible an **on disk** structure. Where it does not matter how many instances you have, there's only one source of truth.
 
Benefit:
- No need to hold an instance of the cache handler `InMemoryFeedStore` for the duration of the app life cycle.


InMemoryFeedStore/InMemoryFeedStore.swift
```diff
@@ -0,0 +1,52 @@
...
+import Foundation
+
+public class InMemoryFeedStore: FeedStore {
+
+	private struct CacheData {
+		let feed: [LocalFeedImage]
+		let timesstamp: Date
+	}
+
+	private static var cache: CacheData?
```
caiozullo, line 18: Since this class is a Singleton, the `cache` property doesn't need to be static. It can be a property of the Singleton instance:

```
private var cache: CacheData?
```

Tests/FeedStoreChallengeTests.swift
```diff
@@ -18,85 +18,112 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
 	//  Repeat this process until all tests are passing.
 	//
 	//  ***********************
+
+	override func setUp() {
+		super.setUp()
+
+		setupEmptyStoreState()
+	}
```
caiozullo, line 26: Missing the `tearDown` method.

If the implementation contains side-effects, it's important to also clean it in `tearDown`. Otherwise, the side-effect will still exist after the test completes. 

Ideally, every test should `start` and `finish` in a clean state.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -20,67 +20,72 @@ class FeedStoreIntegrationTests: XCTestCase {
 		setupEmptyStoreState()
 	}
 	
-	override func tearDown() {
-		super.tearDown()
-		
-		undoStoreSideEffects()
-	}
```
caiozullo, line 27: The `tearDown` should *not* be removed.

If the implementation contains side-effects, it's important to also clean it in `tearDown`. Otherwise, the side-effect will still exist after the test completes. 

Ideally, every test should `start` and `finish` in a clean state.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -20,67 +20,72 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+		wait(for: [exp], timeout: 1.0)
 	}
 	
 	private func undoStoreSideEffects() {
-		
+		let exp = expectation(description: "Cache is cleaned")
+		makeSUT().deleteCachedFeed { (_) in
+			exp.fulfill()
+		}
+		wait(for: [exp], timeout: 1.0)
```
caiozullo, line 89: There's duplication in these two methods. You can use the provided `deleteCache(from: makeSUT())` helper method.

Tests/FeedStoreSpecs/XCTestCase+FeedStoreSpecs.swift
```diff
@@ -10,6 +10,15 @@ extension FeedStoreSpecs where Self: XCTestCase {
...
 	}
+
+	func assertThatRetrieveDeliversFoundValuesOnEmptyInsertedCache(on sut: FeedStore, file: StaticString = #filePath, line: UInt = #line) {
+		let feed = emptyImageFeed()
+		let timestamp = Date()
+
+		insert((feed, timestamp), to: sut)
+
+		expect(sut, toRetrieve: .found(feed: feed, timestamp: timestamp), file: file, line: line)
+	}
```
caiozullo, line 21: This helper is unnecessary - you can remove it. Use the provided tests to validate your implementation.

Tests/FeedStoreSpecs/XCTestCase+FeedStoreHelpers.swift
```diff
@@ -9,6 +9,10 @@ extension XCTestCase {
     func uniqueImageFeed() -> [LocalFeedImage] {
         return [uniqueImage(), uniqueImage()]
     }
+
+	func emptyImageFeed() -> [LocalFeedImage] {
+		return []
+	}
```
caiozullo, line 15: This helper is unnecessary - you can remove it. Use the provided tests to validate your implementation.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -18,85 +18,112 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
-		//		assertThatRetrieveDeliversEmptyOnEmptyCache(on: sut)
+		let sut = makeSUT()
+
+		assertThatRetrieveDeliversEmptyOnEmptyCache(on: sut)
+	}
+
+	func test_retrieve_deliversFoundValuesOnEmptyInsertedCach() {
+		let sut = makeSUT()
+
+		assertThatRetrieveDeliversFoundValuesOnEmptyInsertedCache(on: sut)
```
caiozullo, line 37: This test is unnecessary - you can remove it. Use the provided tests to validate your implementation.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -18,85 +18,112 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
 	}
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT(file: StaticString = #filePath, line: UInt = #line) -> FeedStore {
+		let sut = InMemoryFeedStore.shared
+
+		checkForMemoryLeaks(for: sut as AnyObject)
```
caiozullo, line 111: The casting to `AnyObject` is unnecessary - you can remove it:

```suggestion
		checkForMemoryLeaks(for: sut)
```

Tests/FeedStoreChallengeTests.swift
```diff
@@ -18,85 +18,112 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+	private func setupEmptyStoreState() {
+		undoStoreSideEffects(for: makeSUT())
+	}
+
+	private func undoStoreSideEffects(for sut: FeedStore) {
+		let exp = expectation(description: "Waiting for Cache to be cleaned")
+		sut.deleteCachedFeed { (_) in
+			exp.fulfill()
+		}
+		wait(for: [exp], timeout: 1.0)
```
caiozullo, line 125: You can use the provided `deleteCache(from: makeSUT())` helper method to avoid duplication in this method.

caiozullo (commented):
Hi @halonsoluis, 

caiozullo:
Also, please update the PR name following the instructions in the readme file. Thank you!

Tests/FeedStoreIntegrationTests.swift
halonsoluis: I had the tearDown method, but then, when attempting to track for memory leaks I started getting the following error:

_API violation -- attempting to add a teardown block after teardown blocks have been dequeued (NSGenericException)_

> Caused by the call to `makeSUT()` that is made inside of _tearDown_ when adding the check for memory leaks.

So, as it is an inMemory implementation I weighted my options.

How would you handle this case?

InMemoryFeedStore/InMemoryFeedStore.swift
```diff
@@ -0,0 +1,52 @@
...
+		queue.async(flags: .barrier) {
+			Self.cache = nil
+			completion(nil)
+		}
+	}
+}
+
+extension InMemoryFeedStore {
+	public static var shared: FeedStore {
+		InMemoryFeedStore()
```
halonsoluis, line 50: Now I see.
On my attempt to create a shared helper that would work as a factory I created a singleton!

InMemoryFeedStore/InMemoryFeedStore.swift
halonsoluis: You are totally right, I totally missed the fact that I had created a singleton with the extension below.
My intention was to have a factory method, but I created a singleton instead.

Which approach would you recommend?

Going for a recognisable singleton, or changing the extension to instead provide new instances of this class that share a common static cache?

halonsoluis: I'll go for a plain singleton.
I think that for this case it would make sense, creating other instances just for the sake of it will be counterproductive and would just cause confusion.

halonsoluis: Hi @caiozullo,

I don't think this is a singleton at all, a singleton would require the instance to be held statically as a `let`.
What I have right now on the extension is just a computed variable that would construct new instances of this class.

I went ahead and tested it, and changing the `static` cache into a `non static` one proves it.

The instances are different.

Screenshot 2021-02-10 at 21 35 37

and hence the integration tests dealing with this behaviour start failing

Screenshot 2021-02-10 at 21 36 40

I see where the confusion is generated and I'll go ahead an remove that extra factory, there's no added benefit on it.

halonsoluis: In any case, I liked the idea of using the `.shared` and have a recognisable pattern in place.
So, I'll go along and fix the singleton implementation.

InMemoryFeedStore/InMemoryFeedStore.swift
halonsoluis: Let's ignore this and remove this piece of code as it was a reason for misinterpretation.

Tests/FeedStoreChallengeTests.swift
halonsoluis: It was needed as my `.shared` was returning an instance of the protocol instead of the class to be tested.
This is now addressed and fixed.

InMemoryFeedStore/InMemoryFeedStore.swift
halonsoluis: Hi @caiozullo,

It would make sense to check a `sut` that is a singleton for memory leaks?

If my understanding is correct, it will then be held for the lifetime of the application.

So, I have a dilemma now, revert back into an implementation where new instances are created every time and only the underlying cache is static or remove entirely the code for memory leaks tracking.

Would you mind providing your take on this?

Tests/FeedStoreIntegrationTests.swift
halonsoluis: I can solve this by creating some type of duplication...

Screenshot 2021-02-10 at 22 27 41

Or, as the main goal is not to execute the original `makeSUT()` I could introduce an extra function just for creation of the SUT without any extra checks (like for example, the memory leaks).

Screenshot 2021-02-10 at 22 33 23

InMemoryFeedStore/InMemoryFeedStore.swift
halonsoluis: On commit **[783ef3e](https://github.com/essentialdevelopercom/ios-lead-essentials-feed-store-challenge/pull/82/commits/783ef3eb4cbf0d8ea3f961390b0d3941fd1a1c77)** the singleton approach is used, if `trackMemoryLeaks` is uncommented, all of the test cases will fail due to potential memory leak.

On latest commit **[d1f6fb6](https://github.com/essentialdevelopercom/ios-lead-essentials-feed-store-challenge/pull/82/commits/d1f6fb669a3c153c452545cb24538e0235e478af)** I have reverted to use the multiple instances sharing a common underlying cache data structure held statically in memory as to demonstrate that in this case there's no retention or memory leaks beside the statically held cache data structure.

halonsoluis: I think both approaches should be ok, but it's not up to me to decide here :)

halonsoluis:
The PR is ready to be reviewed again.

Please notice that I went up and down trying to get into a consensus with myself and with your comments. I hope it's not too much rant :)

InMemoryFeedStore/InMemoryFeedStore.swift
caiozullo: Both approaches are ok. 

If you're sharing the cache, you can go with a Singleton. Clients will still interact only with the `FeedStore` protocol, so the Singleton is an implementation detail the clients don't need to know about.

Only the Composition Root will know about the concrete type and access the shared singleton instance.

In the case of a Singleton, you don't test for memory leaks because it's always in memory. But you can add a test to make sure that `.shared` always returns the same instance!

Tests/FeedStoreIntegrationTests.swift
caiozullo: `deleteCache(from: InMemoryFeedStore.shared)` is fine. Moreover, you don't need the memory leak tracking if you make it a Singleton.

caiozullo (commented):
Hi @halonsoluis,

- All tests are passing ✅ 

I replied to your comments - feel free to push changes.

Great job!

InMemoryFeedStore/InMemoryFeedStore.swift
```diff
@@ -0,0 +1,53 @@
...
+//  Created by Hugo Alonso on 05/02/2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+
+public class InMemoryFeedStore: FeedStore {
+	private var cache: CacheData?
+
+	public static let shared: InMemoryFeedStore = InMemoryFeedStore()
```
caiozullo, line 14: The compiler can infer the type:

```suggestion
	public static let shared = InMemoryFeedStore()
```

InMemoryFeedStore/InMemoryFeedStore.swift
```diff
@@ -0,0 +1,53 @@
...
+		label: "\(InMemoryFeedStore.self)Queue",
+		qos: .userInitiated,
+		attributes: .concurrent
+	)
+
+	private init() {}
+
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		queue.async { [weak self] in
+			guard let cache = self?.cache else {
```
caiozullo, line 26: The Singleton instance is always in memory, so there's no need to `weakify` it in this context.

InMemoryFeedStore/InMemoryFeedStore.swift
```diff
@@ -0,0 +1,53 @@
...
+				return completion(.empty)
+			}
+			completion(.found(feed: cache.feed, timestamp: cache.timesstamp))
+		}
+	}
+	
+	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+		queue.async(flags: .barrier) { [weak self] in
+			self?.cache = CacheData(feed: feed, timesstamp: timestamp)
```
caiozullo, line 35: The Singleton instance is always in memory, so there's no need to `weakify` it in this context.

InMemoryFeedStore/InMemoryFeedStore.swift
```diff
@@ -0,0 +1,53 @@
...
+			completion(nil)
+		}
+	}
+
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		queue.async(flags: .barrier) { [weak self] in
+			self?.cache = nil
```
caiozullo, line 42: The Singleton instance is always in memory, so there's no need to `weakify` it in this context.

InMemoryFeedStore/InMemoryFeedStore.swift
halonsoluis: I was wondering about this, good to have confirmation. :)

caiozullo:
Nice one!

PR #84, opened 2021-02-11
Couchbase 2.8 implementation - Ivan Ornes

ivanornes:


Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,56 +27,56 @@ class FeedStoreIntegrationTests: XCTestCase {
...
 	// - MARK: Helpers
 	
 	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
+		return try! CouchFeedStore(databaseName: "feed-database")
 	}
 	
 	private func setupEmptyStoreState() {
-		
+		makeSUT().deleteCachedFeed { _ in }
```
caiozullo, line 79: This operation is async, so it's important to *wait* for it to finish before returning. Otherwise, other tests might start in between - leading to race conditions.

You should use the provided `deleteCache(from: makeSUT())` helper, which will wait for the operation to finish.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,56 +27,56 @@ class FeedStoreIntegrationTests: XCTestCase {
...
 	}
 	
 	private func undoStoreSideEffects() {
```
caiozullo, line 82: This `undoStoreSideEffects` method must also be implemented.

Every test should start and finish in a clean state, to avoid cross-test issues.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,137 +20,83 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
 	}
 	
 	// - MARK: Helpers
 	
 	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
+		let databaseName = "feed-database"
+		let feedStore = try! CouchFeedStore(databaseName: databaseName)
+		feedStore.deleteCachedFeed { _ in }
+		return try! CouchFeedStore(databaseName: databaseName)
```
caiozullo, line 100: The store shouldn't be initiated twice in the `makeSUT`

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,137 +20,83 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		assertThatSideEffectsRunSerially(on: sut)
 	}
 	
 	// - MARK: Helpers
 	
 	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
+		let databaseName = "feed-database"
+		let feedStore = try! CouchFeedStore(databaseName: databaseName)
+		feedStore.deleteCachedFeed { _ in }
```
caiozullo, line 99: Deleting the cache should be done in `setUp` and `tearDown`

Moreover, this operation is async - so it's important to *wait* for it to finish before returning. Otherwise, other tests might start in between - leading to race conditions.

You should use the provided `deleteCache(from: makeSUT())` helper, which will wait for the operation to finish.

FeedStoreChallenge/CouchFeedStore.swift
```diff
@@ -0,0 +1,63 @@
...
+			completion(error)
+		}
+	}
+	
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		let imageFeedQuery = QueryBuilder.allFeedImagesQuery(from: database)
+		do {
+			let mappedImages = try imageFeedQuery.execute().compactMap { $0.localFeedImage }
+			if mappedImages.isEmpty {
+				completion(.empty)
```
caiozullo, line 40: An empty array of images is a valid value in the cache. So this `isEmpty` check shouldn't exist here. (If there's logic related to empty cache, this logic should be somewhere else - not in the store insfrastructure).

It should complete with `.empty` if there are no values stored in the cache, including the `timestamp`.

caiozullo (commented):
Hi @ivanornes, you almost passed the challenge!

There shouldn't be any force-unwrap (`!`) in the production code. I also left review comments.

Let me know once you're done so I can review it again. Thank you 🙂

ivanornes:
Hey @caiozullo! thanks for your constructive feedback! :)

I made the changes and looks much better. I hope now meets the standards 😊

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,137 +20,88 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
 	
 	// - MARK: Helpers
 	
 	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
+		try! CouchFeedStore(databaseName: "feed-database")
+	}
+
+	override func setUp() {
+		deleteCache(from: makeSUT())
```
caiozullo, line 101: It's good practice to call `super.setUp()` when overriding the method, so you're sure the whole class hierarchy setUp is executed (even if the hierarchy changes in the future!).

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,137 +20,88 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
 	}
 	
+	override func tearDown() {
+		deleteCache(from: makeSUT())
```
caiozullo, line 105: It's good practice to call `super.tearDown()` when overriding the method, so you're sure the whole class hierarchy tearDown is executed (even if the hierarchy changes in the future!).

caiozullo (commented):
Hi @ivanornes,

- All tests are passing ✅ 

Great job!

ivanornes:
Updated the setUp and tearDown thanks @caiozullo!

caiozullo:
Nice one!

PR #89, opened 2021-02-16
CoreData implementation - Joe Lauletta

jlappdev:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,65 @@
...
+import Foundation
+import CoreData
+
+public class CoreDataFeedStore: FeedStore {
+	
+	private let context: NSManagedObjectContext
+	
+	public init(withContext context: NSManagedObjectContext) {
+		self.context = context
+	}
```
caiozullo, line 18: This class cannot work with *any* `NSManagedObjectContext` instance. It needs to be an `NSManagedObjectContext` for the right managed model. 

If we pass a context for the wrong model, the instance will be created but will lead to undefined behavior and runtime errors.

So the model, persistent store, and context *should ideally be created internally by this class*. This way, clients cannot pass the wrong context - eliminating runtime issues. It also makes this class easier to use by clients.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,83 +21,164 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
-	private func makeSUT() -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT(file: StaticString = #file, line: UInt = #line) -> FeedStore {
+		let sut = CoreDataFeedStore(withContext: makeCoreDataStack(at: testStoreURL()))
+		trackForMemoryLeaks(sut, file: file, line: line)
+		
+		return sut
 	}
 	
+	private func makeCoreDataStack(at url: URL, using modelName: String = "FeedStoreChallengeModel") -> NSManagedObjectContext {
```
caiozullo, line 104: The `modelName` parameter is never passed to this function. You can remove it.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,83 +21,164 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		let coordinator = NSPersistentStoreCoordinator(managedObjectModel: managedObjectModel)
+		
+		let description = NSPersistentStoreDescription(url: url)
+		description.type = NSSQLiteStoreType
+		
+		coordinator.addPersistentStore(with: description) { (_, error) in
+			if let error = error {
+				fatalError("Unable to create Core Data Stack. Failed with \(error)")
+			}
+		}
```
caiozullo, line 151: This method is almost identical to the `makeCoreDataStack` - there's a lot of duplication that could be removed.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,83 +21,164 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+	
+	private func makeSUTUsingStubbedContext(file: StaticString = #file, line: UInt = #line) -> (sut: FeedStore, context: ManagedObjectContextStub) {
+		let context = makeStubbedCoreDataStack(at: testStoreURL())
+		let sut = CoreDataFeedStore(withContext: context)
+		trackForMemoryLeaks(sut, file: file, line: line)
+		
+		return (sut, context)
+	}
+	
+	private func makeStubbedCoreDataStack(at url: URL, using modelName: String = "FeedStoreChallengeModel") -> ManagedObjectContextStub {
```
caiozullo, line 140: The `modelName` parameter is never passed to this function. You can remove it.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,83 +21,164 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		return mom
+	}
+	
+	private func makeSUTUsingStubbedContext(file: StaticString = #file, line: UInt = #line) -> (sut: FeedStore, context: ManagedObjectContextStub) {
+		let context = makeStubbedCoreDataStack(at: testStoreURL())
+		let sut = CoreDataFeedStore(withContext: context)
+		trackForMemoryLeaks(sut, file: file, line: line)
+		
+		return (sut, context)
+	}
```
caiozullo, line 138: This method is almost identical to the `makeSUT`. You could have only one `makeSUT` but pass the context as a parameter with a default value to eliminate duplication.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +27,102 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+		let managedObjectModel = makeManagedObjectModel()
+		let description = NSPersistentStoreDescription(url: url)
+		let container = NSPersistentContainer(name: modelName, managedObjectModel: managedObjectModel)
+		
+		container.persistentStoreDescriptions = [description]
+		container.loadPersistentStores { (_, error) in
+			if let error = error {
+				fatalError("Unable to create Core Data Stack. Failed with \(error)")
+			}
+		}
```
caiozullo, line 91: The Core Data stack creation is duplicated here as well. Another sign this logic should be inside the `CoreDataFeedStore` class (see my other comment).

caiozullo (commented):
Hi @jlappdev,

- All tests are passing ✅ 

Great job!

jlappdev:
Caio - I really appreciate the feedback from the code review. I have addressed your comments in the new commits using some of the lessons learned from the next video in the course. Unfortunately I was not able to find an elegant way of retaining the failable tests (without leaking test implementation details into the production code) once I moved initialization of the Core Data Stack into 'CoreDataFeedLoader'. As such I had to remove those tests in 8e2b0ac8505b2bd1c4798463022b257303267fcd. I was hoping that digging deeper into the Core Data documentation would reveal an abstract class (similar to how the URL Loading System works) that I could mock to stub errors. But, I was unable to find anything.

caiozullo:
Nice one, @jlappdev!

Regarding the error stub, there's an abstract class (`NSPersistentStore`) you can [register](https://developer.apple.com/documentation/coredata/nspersistentstorecoordinator/1468786-registerstoreclass) similar to the URL Loading System. But unfortunately, it's not very reliable.

PR #92, opened 2021-02-20
InMemory implementation - Nick Karpov

nickkarpov:
One thing is I had failed tests (tracking for memory leaks) when run on GitHub environment - https://github.com/essentialdevelopercom/ios-lead-essentials-feed-store-challenge/pull/92/checks?sha=177ac3623e8a9e883f82e13899c4e3af6b0505be
It is related to capturing self as strong on operationsQueue that owned by self. I wondering why they always pass locally on my machine?


FeedStoreChallenge/InMemoryFeedStore.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+//  Created by Nick on 19/02/2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+
+public class InMemoryFeedStore {
+	private var cachedFeed = [LocalFeedImage]()
+	private var lastUpdate = Date()
```
caiozullo, line 13: The `cachedFeed ` and `lastUpdate` shouldn't have default values.

It'd be best to combine these two values into a new type or a simple tuple. 

This way, you can perform the operation in one step and also represent the `empty` cache with an optional:

```
private var cache: (feed: [LocalFeedImage], timestamp: Date)?
```

FeedStoreChallenge/InMemoryFeedStore.swift
```diff
@@ -0,0 +1,47 @@
...
+			self?.cachedFeed = feed
+			self?.lastUpdate = timestamp
+			completion(nil)
+		}
+	}
+
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		operationsQueue.async { [cachedFeed, lastUpdate] in
+			if cachedFeed.isEmpty {
+				completion(.empty)
```
caiozullo, line 41: An empty array of images is a valid value in the cache - so this class shouldn't check for `isEmpty`. 

This class should complete with `.empty` only when the cache is indeed empty (no value is stored). See my other comment on how to represent it with an optional instead.

caiozullo (commented):
Hi @nickkarpov, you almost passed the challenge.

caiozullo (commented):
Hi @nickkarpov,

- All tests are passing ✅ 
- Bonus points for the Memory Leak Tracking ✅ 

Great job!

caiozullo:
Hi @nickkarpov, regarding your question about the memory leak tracking: 

If the queue closure retains `self` for longer than the test lifetime, it'll fail. But it can fail intermittently since the queue may operation may finish before sometimes, and after the test method finishes other times (the CI serve may be slower, so it may happen there more often). 

So it's important to fix it to have reliable tests regardless of the environment.

nickkarpov:
> Hi @nickkarpov, regarding your question about the memory leak tracking:
> 
> If the queue closure retains `self` for longer than the test lifetime, it'll fail. But it can fail intermittently since the queue may operation may finish before sometimes, and after the test method finishes other times (the CI serve may be slower, so it may happen there more often).
> 
> So it's important to fix it to have reliable tests regardless of the environment.

Thanks Caio,
I also though it could depend on the slower environment on the ci. I just wanted to make sure I can trust my memory leaks tests )

PR #94, opened 2021-02-23
CoreData implementation - Azamat Valitov.

valitovaza:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,110 @@
+import CoreData
+
+public class CoreDataFeedStore: FeedStore {
+	
+	private let modelName = "CoreDataFeedStoreModel"
```
caiozullo, line 5: The `modelName` is only used in the initializer - and it's not required afterward, so it shouldn't be an instance stored property. It could be a `static let` or inlined at the call site, for example.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,110 @@
...
+		try loadError.map { throw $0 }
+		return container
+	}
+}
+
+extension NSOrderedSet {
+	var toLocalFeedImages: [LocalFeedImage] {
+		compactMap({$0 as? CoreDataFeedImage}).compactMap({
+			guard let id = $0.id else { return nil }
+			guard let url = $0.url else { return nil }
```
caiozullo, line 106: Ideally, `id` and `url` shouldn't be optional. So the `guard` statements wouldn't be needed.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,110 @@
...
+		let container = NSPersistentContainer(name: modelName, managedObjectModel: model)
+		container.persistentStoreDescriptions = [description]
+		var loadError: Swift.Error?
+		container.loadPersistentStores { loadError = $1 }
+		try loadError.map { throw $0 }
+		return container
+	}
+}
+
+extension NSOrderedSet {
```
caiozullo, line 102: This extension is only used in this file, so it can be private.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,110 @@
...
+		let request = NSFetchRequest<CoreDataFeed>(entityName: "CoreDataFeed")
+		return try self.context.fetch(request).first
+	}
+}
+
+enum CoreDataFeedStoreError: Error {
+	case invalidModel
+}
+
+extension NSPersistentContainer {
```
caiozullo, line 87: This extension is only used in this file, so it can be private.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +27,73 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+		return FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
 	}
 	
 	private func setupEmptyStoreState() throws {
-		
+		deleteStoreArtifacts()
+	}
+	
+	private func deleteStoreArtifacts() {
+		try? FileManager.default.removeItem(at: testSpecificStoreUrl())
```
caiozullo, line 92: Instead of ignoring errors, this method can throw. If there's something wrong, it'll generate a test failure. This way, it's easier to find and solve issues.

```suggestion
	private func deleteStoreArtifacts() throws {
		try FileManager.default.removeItem(at: testSpecificStoreUrl())
```

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,83 +20,144 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+	
+	func activateNSManagedObjectContextDeleteFailure() {
+		//core data doesn't fail on deletion, it fails on retrieval
+		toggleNSManagedObjectContextRetrieveTestFailure()
+	}
+	
+	func deactivateNSManagedObjectContextDeleteFailure() {
+		//core data doesn't fail on deletion, it fails on retrieval
+		toggleNSManagedObjectContextRetrieveTestFailure()
+	}
```
caiozullo, line 125: Actually, the deletion should fail on the `save` operation.

```suggestion
	func activateNSManagedObjectContextDeleteFailure() {
		//core data doesn't fail on deletion, it fails on save
		toggleNSManagedObjectContextSaveTestFailure()
	}
	
	func deactivateNSManagedObjectContextDeleteFailure() {
		//core data doesn't fail on deletion, it fails on save
		toggleNSManagedObjectContextSaveTestFailure()
	}
```

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +168,70 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		assertThatDeleteDeliversErrorOnDeletionError(on: sut)
+		
+		deactivateNSManagedObjectContextDeleteFailure()
+	}
+
+	func test_delete_hasNoSideEffectsOnDeletionError() throws {
+		let sut = try makeSUT()
+		
+		assertThatDeleteHasNoSideEffectsOnDeletionError(on: sut)
+	}
```
caiozullo, line 235: This test is missing the failure setup:

```suggestion
	func test_delete_hasNoSideEffectsOnDeletionError() throws {
		let sut = try makeSUT()
		
		activateNSManagedObjectContextDeleteFailure()
		
		assertThatDeleteHasNoSideEffectsOnDeletionError(on: sut)
		
		deactivateNSManagedObjectContextDeleteFailure()
	}
```

caiozullo (commented):
Hi @valitovaza,

- All tests are passing ✅ 

Great job!

Moreover, I'd like to point out that `Swizzling` was a good solution to test the error cases in this challenge - but there's a downside: You won't be able to run the tests in parallel, as it'd lead to race conditions.

FeedStoreChallenge/CoreDataFeedStore.swift
valitovaza: Inlined model name

FeedStoreChallenge/CoreDataFeedStore.swift
valitovaza: Made extension `private`

FeedStoreChallenge/CoreDataFeedStore.swift
valitovaza: Made it private

Tests/FeedStoreIntegrationTests.swift
valitovaza: Agree, it's also needed to check if artifacts exist or not

Tests/FeedStoreChallengeTests.swift
valitovaza: Sure it makes sense, I made changes
Thanks

Tests/FeedStoreChallengeTests.swift
valitovaza: Added failure setup, the previous deletion error simulation helped this test pass. There was a problem with the delete error simulation.

FeedStoreChallenge/CoreDataFeedStore.swift
valitovaza: Remove code generation for CoreDataFeedImage to remove optional properties

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,108 @@
...
+		try loadError.map { throw $0 }
+		return container
+	}
+}
+
+private extension NSOrderedSet {
+	var toLocalFeedImages: [LocalFeedImage] {
+		compactMap({$0 as? CoreDataFeedImage}).compactMap({
+			return LocalFeedImage(id: $0.id, description: $0.desc, location: $0.location, url: $0.url)
+		})
```
caiozullo, line 106: Since the return is not optional anymore, you can use `map` instead:

```suggestion
		compactMap({$0 as? CoreDataFeedImage}).map({
			return LocalFeedImage(id: $0.id, description: $0.desc, location: $0.location, url: $0.url)
		})
```

caiozullo (commented):
Nice one!

PR #98, opened 2021-02-27
Realm implementation - Salah Nahed

salahamassi:


FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,134 @@
...
+//  Created by Salah Amassi on 27/02/2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+public class RealmFeedStore: FeedStore {
+	
+	let storeURL: URL?
```
caiozullo, line 14: This property should be private.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,134 @@
...
+			if let cache = realm.objects(Cache.self).last {
+				completion(.found(feed: cache.feed, timestamp: cache.timestamp))
+			} else {
+				completion(.empty)
+			}
+		} catch Realm.Error.fileNotFound, Realm.Error.filePermissionDenied {
+			completion(.empty)
+		} catch {
+			completion(.failure(error))
+		}
```
caiozullo, line 32: If there's *any* error, the store should complete with `completion(.failure(error))`.

```suggestion
		} catch {
			completion(.failure(error))
		}
```

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,134 @@
...
+	}
+	
+	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+		do {
+			let realm = try getRealm()
+			try realm.write {
+				realm.add(Cache(realmFeed: feed.toRealmList(), timestamp: timestamp))
```
caiozullo, line 39: Since we're always *overriding* the existing cache, the implementation should delete the existing cache *before* inserting the new values. Otherwise, it'll use too much unnecessary disk space.

```suggestion
				realm.deleteAll()
				realm.add(Cache(realmFeed: feed.toRealmList(), timestamp: timestamp))
```

Tests/FeedStoreChallengeTests.swift
```diff
@@ -19,84 +19,117 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		assertThatSideEffectsRunSerially(on: sut)
 	}
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() throws -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT(url: URL? = nil) throws -> FeedStore {
+		let store = RealmFeedStore(storeURL: url ?? testSpecificStoreURL())
+		trackForMemoryLeaks(store)
```
caiozullo, line 110: Missing file/line parameters to show test failures in the right place.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +27,73 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+
+		expect(storeToLoad, toRetrieve: .empty)
 	}
 	
 	// - MARK: Helpers
 	
 	private func makeSUT() throws -> FeedStore {
-		fatalError("Must be implemented")
+		let store = RealmFeedStore(storeURL: testSpecificStoreURL())
+		trackForMemoryLeaks(store)
```
caiozullo, line 76: Missing file/line parameters to show test failures in the right place.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -19,84 +19,117 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+
+		assertThatSideEffectsRunSerially(on: sut)
 	}
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() throws -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT(url: URL? = nil) throws -> FeedStore {
+		let store = RealmFeedStore(storeURL: url ?? testSpecificStoreURL())
```
caiozullo, line 109: You can use an in-memory Realm store to run these tests. It's faster and doesn't require you to clean up the artifacts saved to disk. 

So instead of injecting a `URL` in the initializer, you could inject the `RealmConfiguration`. And it's up to the clients to decide to use a persistent config or an in-memory one.

The integration tests should use a real disk URL to test the persistence end-to-end. But the tests in this file don't need it.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +140,57 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
-//	}
-//
-//	func test_retrieve_hasNoSideEffectsOnFailure() throws {
-////		let sut = try makeSUT()
-////
-////		assertThatRetrieveHasNoSideEffectsOnFailure(on: sut)
-//	}
-//
-//}
+extension FeedStoreChallengeTests: FailableRetrieveFeedStoreSpecs {
```
caiozullo, line 143: The setup used for all failable tests in this file aren't fully testing for the specified scenarios because the store can't even be created with a URL pointing to a non-realm DB file. Moreover, some tests have been modified and are using the wrong assertion helper.

There's no easy to simulate Realm errors - and it can complicate the design (by adding protocols that can be mocked, for example). So for this challenge, you can skip these tests and remove the failable specs.

But if you want to do it, you should find a different way of simulating errors with Realm and use the correct assertions. You can check the other PRs to learn how other students did it.

caiozullo (commented):
Hi @salahamassi, you almost passed the challenge!

salahamassi:
Hello @caiozullo 
please review it again
and thanks!
and can I use mocking to test FailableFeedStoreSpecs? or this not including this challenge?

caiozullo:
Hi @salahamassi, you missed some of my comments. Please, check the comments I haven't marked as resolved.

You can use mocking if you want to practice, but it's not needed for this challenge. There are other solutions too, like using read-only stores instead of mocking. Check how other students did it if you want to practice.

salahamassi:
> resolved

thank you a lot, I will add FailableFeedStoreSpecs for practice at the weekend.

caiozullo (commented):
Hi @salahamassi,

- All tests are passing ✅ 

Great job!

PR #102, opened 2021-02-28
InMemory implementation - Kubilay Erdogan

kublaios:


FeedStoreChallenge/InMemoryFeedStore.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+public final class InMemoryFeedStore {
+	private struct Cache {
+		let feed: [LocalFeedImage]
+		let timestamp: Date
+	}
+
+	public static let shared = InMemoryFeedStore.init()
```
caiozullo, line 9: The shared instance is not needed. You can remove it.

Use a public initializer instead, so clients can decide if they want to share instances or not.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -6,6 +6,18 @@ import XCTest
...
+		super.setUp()
+
+		self.deleteCache(from: self.makeSUT())
+	}
+
+	override func tearDown() {
+		super.tearDown()
+
+		self.deleteCache(from: self.makeSUT())
+	}
```
caiozullo, line 20: Deleting the cache on setup/teardown is not necessary if the instance is not `shared` - you can remove these methods (see my other comments).

caiozullo (commented):
Hi @kublaios,

- All tests are passing ✅ 

Great job!

kublaios:
Thanks for the review, @caiozullo! Good point with the shared instance. Refactored the code accordingly.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -19,138 +19,92 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		assertThatSideEffectsRunSerially(on: sut)
 	}
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() throws -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT() -> FeedStore {
+		let sut = InMemoryFeedStore.init()
+		self.trackForMemoryLeaks(sut)
```
caiozullo, line 98: Missing file/line params to show errors in the right place.

Also, `self.` is not needed in this context.

```swift
  trackForMemoryLeaks(sut, file: file, line: line)
```

Tests/FeedStoreChallengeTests.swift
```diff
@@ -19,138 +19,92 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+
+		assertThatSideEffectsRunSerially(on: sut)
 	}
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() throws -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT() -> FeedStore {
+		let sut = InMemoryFeedStore.init()
```
caiozullo, line 97: `.init` is unnecessary use `InMemoryFeedStore()` instead.

Tests/FeedStoreChallengeTests.swift
kublaios: Yes, I still haven't used to append the file and line info in the parent methods on the call trace... Fixed that, thanks.

As for the explicit calls to `self`, it's my personal preference since I believe it makes the scope clearer (especially during code reviews) but since it's the convention of the host project, I removed it here to follow the conventions :)

caiozullo:
Nice one!

PR #106, opened 2021-03-06
Couchbase Lite implementation - Fabio Mignogna

migfabio:
Unfortunately [Couchbase Lite does not support in-memory database](https://forums.couchbase.com/t/can-couchbase-lite-be-configured-as-an-in-memory-database/5688/6), so I had to create artefacts also for the unit tests.


FeedStoreChallenge/CouchbaseLiteFeedStore.swift
```diff
@@ -0,0 +1,153 @@
...
+				  let urlString = json.string(forKey: "url"), let url = URL(string: urlString) else {
+				return nil
+			}
+
+			self.id = id
+			self.description = json.string(forKey: "description")
+			self.location = json.string(forKey: "location")
+			self.url = url
+		}
+	}
```
caiozullo, line 80: This private struct is identical to the `LocalFeedImage` so all its methods could be just private extensions on the `LocalFeedImage`.

FeedStoreChallenge/CouchbaseLiteFeedStore.swift
```diff
@@ -0,0 +1,153 @@
...
+
+	private lazy var database: Database? = {
+		let dbConfig = DatabaseConfiguration()
+		dbConfig.directory = storeURL.path
+		return try? Database(name: "feed-store", config: dbConfig)
+	}()
```
caiozullo, line 86: The `database` should be optional because this class cannot work without one.

Error shouldn't be ignored. They should be handled accordingly to avoid runtime issues.

FeedStoreChallenge/CouchbaseLiteFeedStore.swift
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+		guard let database = database else {
+			return completion(nil)
+		}
+
+		queue.async(flags: .barrier) {
+			do {
+				try database.saveDocument(Cache(feed: feed, timestamp: timestamp).toDocument)
```
caiozullo, line 125: It should delete any existing document before saving the new one. Or will `saveDocument` override existing values?

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -7,80 +7,86 @@ import FeedStoreChallenge
...
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() throws -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT(file: StaticString = #filePath, line: UInt = #line) throws -> FeedStore {
+		let sut = CouchbaseLiteFeedStore(storeURL: storeURL())
+		addTeardownBlock { [weak sut] in
+			XCTAssertNil(sut, file: file, line: line)
+		}
```
caiozullo, line 71: This logic is duplicated in two test files. It should be a shared helper function to avoid duplication.

caiozullo (commented):
Hi @migfabio, you almost passed the challenge!

Please, check my review comments and let me know once you're done.

Thank you 🙂

FeedStoreChallenge/CouchbaseLiteFeedStore.swift
migfabio: I pass a fixed ID to the Document, so when you save the document with the same ID, it will override. I extract the document ID into a static property to clarify intent

migfabio:
Thanks @caiozullo for your feedback. I've addressed all of them 👍

FeedStoreChallenge/CouchbaseLiteFeedStore.swift
```diff
@@ -0,0 +1,131 @@
...
+				"description": description,
+				"location": location,
+				"url": url.absoluteString
+			].compactMapValues { $0 }
+		)
+	}
+
+	var toLocalFeedImage: LocalFeedImage {
+		LocalFeedImage(id: id, description: description, location: location, url: url)
+	}
```
caiozullo, line 118: This method is not needed because it doesn't do anything - It's converting a `LocalFeedImage` into a `LocalFeedImage`.

caiozullo (commented):
Hi @migfabio,

- All tests are passing ✅ 

I left one review comment. Feel free to push changes. 

Great job!

FeedStoreChallenge/CouchbaseLiteFeedStore.swift
migfabio: Done! Thank you very much for your feedbacks 🙌

caiozullo:
Nice one!

PR #109, opened 2021-03-08
CoreData implementation - Edgar Hirama

edhirama:


Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +27,70 @@ class FeedStoreIntegrationTests: XCTestCase {
...
 	private func setupEmptyStoreState() throws {
-		
+
 	}
-	
+
 	private func undoStoreSideEffects() throws {
-		
+
 	}
```
caiozullo, line 86: It's important to delete the artifacts stored to the `testStoreURL` here - every test should start and finish in a clean state to avoid influencing the result of other tests.

FeedStoreChallenge/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,131 @@
...
+extension CoreDataFeedStore {
+
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		let managedContext = context
+		do {
+			let fetchRequest = NSFetchRequest<LocalCache>(entityName: "LocalCache")
+
+			let result = try managedContext.fetch(fetchRequest)
+			if let cacheResult = result.first {
+				completion(.found(feed: cacheResult.feed.array.compactMap { ($0 as? CacheFeedImage)?.localFeedImage } , timestamp: cacheResult.timestamp))
```
caiozullo, line 68: You can create a `localFeed` helper to simplify this line here:

```suggestion
				completion(.found(feed: cacheResult.localFeed, timestamp: cacheResult.timestamp))
```

caiozullo (commented):
Hi @edhirama, you almost passed the challenge!

 Thank you 🙂

edhirama:
Hi @caiozullo, thank you for your review and help :)

I have made the changes you've suggested, but I have a few questions:
How would be a way to test that perform is actually needed, since all tests were passing without it? 
Another topic would be the `deleteStoreArtifacts` method not actually making a difference to make my integration tests pass, that's why I've missed it. 
I checked and the tests are running in random order, so I'd expect to see a failing test when checking for an empty cache some time, but I never see it. Do you know why that is happening in my instance?

I have also tried to remove the randomizing and move the empty check after a test that I knew would leave artifacts behind, but it still didn't fail. :/

FeedStoreChallenge/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,130 @@
...
+		}
+		if let error = loadError {
+			throw CoreDataError.unableToLoad(error)
+		}
+		return container
+	}
+
+	private func deleteCurrentCacheIfNeeded(in context: NSManagedObjectContext) {
+		let fetchRequest: NSFetchRequest<LocalCache> = NSFetchRequest(entityName: LocalCache.className())
+		if let fetchResult = try? context.fetch(fetchRequest), let existingFeed = fetchResult.first {
```
caiozullo, line 49: This line can be simplified in a single statement:

```suggestion
		if let existingFeed = try? context.fetch(fetchRequest).first {
```

FeedStoreChallenge/CoreDataFeedStore/MappedModels.swift
```diff
@@ -0,0 +1,34 @@
...
+@objc(LocalCache)
+public class LocalCache: NSManagedObject {
+
+	@NSManaged public var feed: NSOrderedSet
+	@NSManaged public var timestamp: Date
+
+	public var localFeed: [LocalFeedImage] {
+		feed.array.compactMap { ($0 as? CacheFeedImage)?.localFeedImage }
+	}
+}
```
caiozullo, line 21: This class and its properties shouldn't be public. They're only used internally.

FeedStoreChallenge/CoreDataFeedStore/MappedModels.swift
```diff
@@ -0,0 +1,34 @@
...
+
+	@NSManaged public var id: UUID
+	@NSManaged public var imageDescription: String?
+	@NSManaged public var location: String?
+	@NSManaged public var url: URL
+
+	public var localFeedImage: LocalFeedImage {
+		LocalFeedImage(id: id, description: imageDescription, location: location, url: url)
+	}
+}
```
caiozullo, line 34: This class and its properties shouldn't be public. They're only used internally.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +27,80 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+	}
+
+	private func testStoreURL() -> URL {
+		cacheDirectory().appendingPathComponent("\(String(describing: self)).store")
+	}
+
+	private func trackForMemoryLeaks(_ instance: AnyObject, file: StaticString = #filePath, line: UInt = #line) {
+		addTeardownBlock { [weak instance] in
+			XCTAssertNil(instance, file: file, line: line)
+		}
```
caiozullo, line 104: This method is duplicated in both test files. You can extract it into a separate file and share it to avoid duplication.

Moreover, you can add an assertion message providing more context to make it easier to understand test failures.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +27,80 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+	private func deleteStoreArtifacts() {
+		try? FileManager.default.removeItem(at: testStoreURL())
+	}
+
+	private func cacheDirectory() -> URL {
+		FileManager.default.urls(for: .cachesDirectory, in: .localDomainMask).first!
+	}
+
+	private func testStoreURL() -> URL {
+		cacheDirectory().appendingPathComponent("\(String(describing: self)).store")
```
caiozullo, line 98: `String(describing: self)` will generate an instance-specific string. Every instance has a different `id`, so every test method will have a different `testStoreURL`.

That's why you didn't have to clean the cache for the tests to pass. But it's important to clear the cache to avoid leaving side-effects on the disk.

You can use `type(of: self)` instead, which always returns the same string.

```suggestion
		cacheDirectory().appendingPathComponent("\(type(of: self)).store")
```

caiozullo (commented):
Hi @edhirama, almost there!

Please, check my new review and let me know once you're done so I can review it again.

Regarding the object context threading model, you can write tests forcing race conditions to make sure it's correct. You can also set the argument `-com.apple.CoreData.ConcurrencyDebug 1` in your scheme configuration to enable debug logs!

FeedStoreChallenge/CoreDataFeedStore/MappedModels.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+
+import Foundation
+import CoreData
+
+@objc(LocalCache)
+internal class LocalCache: NSManagedObject {
+
+	@NSManaged public var feed: NSOrderedSet
+	@NSManaged public var timestamp: Date
```
caiozullo, line 16: Those properties shouldn't be public.

FeedStoreChallenge/CoreDataFeedStore/MappedModels.swift
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+@objc(CacheFeedImage)
+internal class CacheFeedImage: NSManagedObject {
+
+	@NSManaged public var id: UUID
+	@NSManaged public var imageDescription: String?
+	@NSManaged public var location: String?
+	@NSManaged public var url: URL
```
caiozullo, line 29: Those properties shouldn't be public.

Tests/XCTestCase+Helpers.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Edgar Hirama on 09/03/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import XCTest
+
+extension XCTestCase {
+	func trackForMemoryLeaks(_ instance: AnyObject, file: StaticString = #filePath, line: UInt = #line) {
+		addTeardownBlock { [weak instance] in
+			XCTAssertNil(instance, file: file, line: line)
```
caiozullo, line 14: As shown in the lectures, you can add an assertion message providing more context to make it easier to understand test failures.

caiozullo (commented):
Hi @edhirama, 

- All tests are passing ✅ 

Great job!

caiozullo:
Nice one!

PR #111, opened 2021-03-10
Realm implementation - Romain Brunie

rombiddle:


caiozullo:
Hi @rombiddle, the CI checks have failed. Please, have a look at let me know when it's fixed. Thank you 😊

caiozullo:
Hi @rombiddle, please use SPM so there aren't so many files in the diff. Otherwise, it'll be very hard to review!

You can check how other students did it. For example: https://github.com/essentialdevelopercom/ios-lead-essentials-feed-store-challenge/pull/98

Thank you!

rombiddle:
Hi @caiozullo , all tests passed!

FeedStoreChallenge.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
caiozullo, line 1: You can remove the workspace since you're not cocoapods anymore.

FeedStoreChallenge/RealmFeedCache.swift
```diff
@@ -0,0 +1,30 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Romain Brunie on 09/03/2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+public class RealmFeedCache: Object {
```
caiozullo, line 12: This class and all its properties/methods shouldn't be public. They're only used internally.

FeedStoreChallenge/RealmFeedImage.swift
```diff
@@ -0,0 +1,36 @@
...
+		self.init()
+		self.id = id.uuidString
+		self.desc = description
+		self.location = location
+		self.url = url.absoluteString
+	}
+	
+	public func toLocal() throws -> LocalFeedImage {
+		guard let uuid = UUID(uuidString: self.id), let url = URL(string: self.url) else {
+			throw NSError(domain: "any error", code: 0)
```
caiozullo, line 28: Provide more context in the error, instead of just "any error".

You can create a custom error type such as `struct InvalidFeedImageData: Error {}`. Then, you can throw it:

```swift
throw InvalidFeedImageData()
```

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,76 @@
...
+				completion(error)
+			}
+		}
+	}
+	
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		queue.async {
+			guard let realm = try? Realm(configuration: self.configuration) else {
+				return completion(.empty)
+			}
```
caiozullo, line 61: Errors shouldn't be ignored with `try?` here. 

If there an error creating the Realm instance, it should complete with failure (not .empty).

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,83 +33,105 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+
+		assertThatSideEffectsRunSerially(on: sut)
 	}
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() throws -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT(configuration: Realm.Configuration? = nil, file: StaticString = #filePath, line: UInt = #line) throws -> FeedStore {
+		let sut = RealmFeedStore(configuration: configuration ?? testRealmConfiguration())
```
caiozullo, line 110: You can use an in-memory Realm configuration to speed up the tests here - you also wouldn't need to clear the file system because it won't persist anything.

The integration tests should run with the real URL to test the whole integration with the file system. But the tests in this file don't need it.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,83 +33,105 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		trackForMemoryLeaks(sut)
```
caiozullo, line 111: Missing file/line params to show test failures in the right places.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +28,62 @@ class FeedStoreIntegrationTests: XCTestCase {
...
+		expect(storeToLoad, toRetrieve: .empty)
 	}
 	
 	// - MARK: Helpers
 	
-	private func makeSUT() throws -> FeedStore {
-		fatalError("Must be implemented")
+	private func makeSUT(configuration: Realm.Configuration? = nil, file: StaticString = #filePath, line: UInt = #line) throws -> FeedStore {
+		let sut = RealmFeedStore(configuration: configuration ?? testRealmConfiguration())
+		trackForMemoryLeaks(sut)
```
caiozullo, line 77: Missing file/line params to show test failures in the right places.

caiozullo (commented):
Hi @rombiddle, you almost passed the challenge!

Thank you 🙂

Tests/FeedStoreChallengeTests.swift
rombiddle: @caiozullo , If I use in-memory realm conf, I will not be able to test realm file that have no write permission. This is the reason why I used real URL.

FeedStoreChallenge.xcodeproj/xcshareddata/xcschemes/FeedStoreChallenge.xcscheme
```diff
@@ -50,6 +50,7 @@
       <Testables>
          <TestableReference
             skipped = "NO"
+            parallelizable = "YES"
```
caiozullo, line 53: Please, undo changes to the scheme - that's not part of the challenge.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,75 @@
...
+					let feed = try conf.realmFeedtoLocals()
+					completion(.found(feed: feed, timestamp: conf.timestamp))
+				} else {
+					completion(.empty)
+				}
+			} catch is RealmFeedImage.InvalidFeedImageData {
+				completion(.failure(RealmFeedImage.InvalidFeedImageData()))
+			} catch {
+				completion(.empty)
+			}
```
caiozullo, line 72: In the case of *any* error, it should complete with `completion(.failure(error))`:

```suggestion
			} catch {
				completion(.failure(error))
			}
```

If the tests are not passing with this logic, the test setup is wrong. Review the test side.

Tests/FeedStoreChallengeTests.swift
caiozullo: You can use a configuration with an invalid URL only for those tests that need it. The others can run with in-memory configurations.

FeedStoreChallenge/RealmFeedStore.swift
rombiddle: Hi @caiozullo , if the configuration url to create Realm is invalid, then I cannot create the Realm instance and I should return empty like the test wants and it is what we did in the course. Can you give me any hint to fix this issue ?

FeedStoreChallenge/RealmFeedStore.swift
rombiddle: hi @caiozullo , If i remove the catch case RealmFeedImage.InvalidFeedImageData, then I have no idea how to fix it because if the configuration url to create Realm is invalid, then I cannot create the Realm instance and I should return empty like the test wants and it is what we did in the course. I am confused with the way to fix this issue

FeedStoreChallenge/RealmFeedStore.swift
caiozullo: Hi @rombiddle, if it fails to create the Realm instance, it should complete with failure (not .empty) because that's an error.

So using an invalid URL is not simulating a 'retrieval error' with Realm. It's simulating a Realm instantiation error. 

In the course we were not using Realm, so you need to find a way to simulate the 'retrieval error' with Realm. If you can't simulate a Realm error, I recommend deleting this test.

FeedStoreChallenge/RealmFeedStore.swift
caiozullo: Hi @rombiddle, if it fails to create the Realm instance, it should complete with failure (not .empty) because that's an error.

So using an invalid URL is not simulating a 'retrieval error' with Realm. It's simulating a Realm instantiation error. 

In the course we were not using Realm, so you need to find a way to simulate the 'retrieval error' with Realm. If you can't simulate a Realm error, I recommend deleting this test.

rombiddle: Hi @caiozullo , I understand I should not return .empty but I feel like if don't handle the noDeletePermissionURL and invalidStoreURL cases for Realm, I am not testing all error cases. For me, testing means I should test everything that can fails, so my app can never crash. In this case if I delete FailableInsertFeedStoreSpecs and FailableDeleteFeedStoreSpecs tests, I am not testing everything.

caiozullo: Hi @rombiddle, that's right - you should test all scenarios you can. But the test setup you had in there was not testing the behavior you want. 

If you want to test all the scenarios with `Realm`, check other PRs here to see how other students did it.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,75 @@
...
+					let feed = try conf.realmFeedtoLocals()
+					completion(.found(feed: feed, timestamp: conf.timestamp))
+				} else {
+					completion(.empty)
+				}
+			} catch Realm.Error.filePermissionDenied, Realm.Error.fileAccess {
+				completion(.empty)
+			} catch {
+				completion(.failure(error))
+			}
```
caiozullo, line 72: To pass the challenge, you need to remove the first `catch`.

Any error should complete with `failure`.

```suggestion
			} catch {
				completion(.failure(error))
			}
```

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,83 +33,117 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		deleteStoreArtifacts()
 	}
 	
+	private func deleteStoreArtifacts() {
+		let realm = try! Realm(configuration: testRealmInMemoryConfiguration())
+		
+		try! realm.write {
+			realm.deleteAll()
+		}
+	}
```
caiozullo, line 146: You don't need to delete stored artifacts because you're using an in-memory config.

An in-memory config doesn't store artifacts.

Remove these methods (including the setUp/tearDown).

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,75 @@
...
+			do {
+				let realm = try Realm(configuration: self.configuration)
+				try realm.write {
+					let feed = feed.map { RealmFeedImage(id: $0.id,
+														 description: $0.description,
+														 location: $0.location,
+														 url: $0.url)
+					}
+					let realmConf = RealmFeedCache(feed: feed, timestamp: timestamp)
+					realm.add(realmConf as Object, update: .modified)
```
caiozullo, line 48: The casting is not needed.

```suggestion
					realm.add(realmConf, update: .modified)
```

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,75 @@
...
+		queue.async(flags: .barrier) {
+			do {
+				let realm = try Realm(configuration: self.configuration)
+				try realm.write {
+					let feed = feed.map { RealmFeedImage(id: $0.id,
+														 description: $0.description,
+														 location: $0.location,
+														 url: $0.url)
+					}
+					let realmConf = RealmFeedCache(feed: feed, timestamp: timestamp)
```
caiozullo, line 47: What does `realmConf` mean? You could rename it to clarify - something like `let cache = `

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,75 @@
...
+	public init(configuration: Realm.Configuration) {
+		self.configuration = configuration
+	}
+	
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		queue.async(flags: .barrier) {
+			do {
+				let realm = try Realm(configuration: self.configuration)
+				let confObject = realm.objects(RealmFeedCache.self).first
+				if let conf = confObject {
```
caiozullo, line 25: What do `confObject` and `conf` mean? You can rename them to clarify.

FeedStoreChallenge/RealmFeedStore.swift
```diff
@@ -0,0 +1,75 @@
...
+			}
+		}
+	}
+	
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		queue.async {
+			do {
+				let realm = try Realm(configuration: self.configuration)
+				let realmObject = realm.objects(RealmFeedCache.self).first
+				if let conf = realmObject {
```
caiozullo, line 62: What's `conf`? You can rename it to clarify.

rombiddle:
hi @caiozullo , I am done

caiozullo (commented):
Hi @rombiddle, 

- All tests are passing ✅ 

Great job!

PR #114, opened 2021-03-15
InMemoryFeedStore - Fernando Cardenas

fernandocardenasm:


FeedStoreChallenge/FeedStore.swift
```diff
@@ -27,3 +27,33 @@ public protocol FeedStore {
 	/// Clients are responsible to dispatch to appropriate threads, if needed.
 	func retrieve(completion: @escaping RetrievalCompletion)
 }
+
+public final class InMemoryFeedStore: FeedStore {
```
caiozullo, line 31: The `InMemoryFeedStore` implementation should be moved to its own file.

caiozullo (commented):
Hi @fernandocardenasm, 

- All tests are passing ✅ 

I left a review comment. Feel free to push changes. 

Great job!

FeedStoreChallenge/FeedStore.swift
```diff
@@ -27,3 +27,33 @@ public protocol FeedStore {
...
+	
+	private struct Cache {
+		let feed: [LocalFeedImage]
```
fernandocardenasm, line 34: @caiozullo Thanks for the feedback!

Just one short question: Would you also agree that in this case was better to use "LocalFeedImage" here, instead of a private "CacheFeedImage", due to the fact that it doesn't need to conform to any protocol or infrastructure class?

caiozullo: Hi @fernandocardenasm. Yes, you should use the `LocalFeedImage` in this case!

PR #119, opened 2021-03-23
CoreData implementation - Daniel Moro

danielmoro:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,122 @@
...
+	
+	public init(storeURL: URL, modelBundle: Bundle = Bundle.main) throws {
+		guard let modelURL = modelBundle.url(forResource: CoreDataFeedStore.modelName, withExtension: "momd"),
+			  let model = NSManagedObjectModel(contentsOf: modelURL) else {
+			throw Error.modelNotFound
+		}
+		
+		container = NSPersistentContainer(name: CoreDataFeedStore.modelName, managedObjectModel: model)
+		container.persistentStoreDescriptions = [NSPersistentStoreDescription(url: storeURL)]
+		container.loadPersistentStores(completionHandler: {_,_ in})
```
caiozullo, line 28: `loadPersistentStores` may fail, but the error is being ignored inside the closure with `_`.

If there's an error, it should be thrown so clients can handle it. Otherwise, it'll lead to runtime issues.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,122 @@
...
+		context = container.newBackgroundContext()
+	}
```
caiozullo, line 30: The initializer is quite long. You can extract the setup in helper methods as shown in the lectures.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,122 @@
...
+		id = image.id
+		imageDescription = image.description
+		location = image.location
+		url = image.url
+	}
+	
+	var local: LocalFeedImage? {
+		guard let id = id, let url = url else {
+			return nil
+		}
```
caiozullo, line 90: `id` and `url` shouldn't be optional. So this `guard` statement wouldn't be needed.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,122 @@
...
+			} catch {
+				completion(error)
+			}
+		}
+	}
+	
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		perform { context in
+			do {
+				if let cache = try CoreDataCache.fetch(in: context), let timestamp = cache.date {
```
caiozullo, line 60: `cache.date` shouldn't be optional - so the `if let timestamp` wouldn't be needed.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,83 +20,101 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
 	}
 	
 	// - MARK: Helpers
 	
 	private func makeSUT() throws -> FeedStore {
-		fatalError("Must be implemented")
+		let bundle = Bundle(for: CoreDataFeedStore.self)
+		let storeURL = URL(fileURLWithPath: "/dev/null")
+		let sut = try CoreDataFeedStore(storeURL: storeURL, modelBundle: bundle)
+		trackMemoryLeaks(sut)
```
caiozullo, line 100: Missing file/line params to show test failures in the right places.

Tests/FeedStoreChallengeTests.swift
```diff
@@ -20,83 +20,101 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		return sut
 	}
 	
+	private func trackMemoryLeaks(
```
caiozullo, line 104: This method is duplicated in both test files. You should extract it to another file so you can share it.

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +27,93 @@ class FeedStoreIntegrationTests: XCTestCase {
...
 	}
 	
 	// - MARK: Helpers
 	
 	private func makeSUT() throws -> FeedStore {
-		fatalError("Must be implemented")
+		let bundle = Bundle(for: CoreDataFeedStore.self)
+		let storeURL = specificStoreURL()
+		let sut = try CoreDataFeedStore(storeURL: storeURL, modelBundle: bundle)
+		trackMemoryLeaks(sut)
```
caiozullo, line 78: Missing file/line params to show test failures in the right places.

caiozullo (commented):
Hi @danielmoro, you almost passed the challenge!

FeedStoreChallenge/CoreDataFeedStore.swift
danielmoro: Hi Caio,
I know that those two fields should not be optional. However, even if I marked them as not optional in the model file, the compiler still generates classes with those attributes as optional. I saw in the lecture after the challenge that You manually generate code for Managed Objects, do you think that this is a viable option just to have attributes of the wanted type?

FeedStoreChallenge/CoreDataFeedStore.swift
danielmoro: This error can be thrown later when the object is already initialized. I have thoughts to block this call with semaphore just to be able to fail if the error is raised but later I decided not to do that. Another option was to save error as a state in some enum:
```
enum containerState {
    case notLoaded
    case loading
    case loaded
    case failed(Error)
}
```
and then tho check this every time in `perform`

Do you have some suggestions?

FeedStoreChallenge/CoreDataFeedStore.swift
caiozullo: Hi @danielmoro, yes, you should define the attributes your code needs as much as you can. This way, you don't need to deal with incompatible APIs, optionals, etc.

FeedStoreChallenge/CoreDataFeedStore.swift
caiozullo: @danielmoro the error shouldn't be thrown later because this object cannot work without a persistent store. It's unusable at runtime without one. So it shouldn't be possible to instantiate it with the persistent store.

You don't need to block the call because it's synchronous by default (even though it's closure-based). We show how to do it in the lectures - you can also check how other students did it in other pull requests ✅

danielmoro:
Hi, @caiozullo thank you for your helpful comments. I have made suggested changes and pushed new commits.

FeedStoreChallenge/CoreDataCache.swift
```diff
@@ -0,0 +1,47 @@
...
+//  
+//
+
+import Foundation
+import CoreData
+
+class CoreDataCache: NSManagedObject {
+	
+	@NSManaged var date: Date
+	@NSManaged var feed: NSOrderedSet?
```
caiozullo, line 13: `feed` shouldn't be optional

FeedStoreChallenge/CoreDataFeedImage.swift
```diff
@@ -0,0 +1,31 @@
...
+extension CoreDataFeedImage {
+	convenience init(_ image: LocalFeedImage, insertInto context: NSManagedObjectContext) {
+		self.init(context: context)
+		id = image.id
+		imageDescription = image.description
+		location = image.location
+		url = image.url
+	}
+	
+	var local: LocalFeedImage? {
```
caiozullo, line 28: The `local` representation can always be created, so the return shouldn't be optional.

```suggestion
	var local: LocalFeedImage {
```

Tests/FeedStoreIntegrationTests.swift
```diff
@@ -27,60 +27,77 @@ class FeedStoreIntegrationTests: XCTestCase {
...
-		fatalError("Must be implemented")
+	private func makeSUT(file: StaticString = #filePath, line: UInt = #line) throws -> FeedStore {
+		let bundle = Bundle(for: CoreDataFeedStore.self)
+		let storeURL = specificStoreURL()
+		let sut = try CoreDataFeedStore(storeURL: storeURL, modelBundle: bundle)
+		trackMemoryLeaks(sut, file: file, line: line)
+		return sut
+	}
+	
+	private func specificStoreURL() -> URL {
```
caiozullo, line 82: You could rename this method. In the lectures, we call it `testSpecificStoreURL` to clarify the context.

caiozullo (commented):
Hi @danielmoro, 

- All tests are passing ✅ 

Great job!

caiozullo:
Nice one!

PR #121, opened 2021-03-28
ObjectBox implementation - Riccardo Rossi

RRReDz:
Hi Caio & Mike,
Since I've already used Realm and the other solutions (except for Core Data, but I saw that the next lectures you will cover that topic), I tried something new, just to experiment! I hope it's fine for you and will not be too complicated to review, if so, I will re-implement everything with one of the suggested solutions.

Just few info:
1. The in-memory storage is not available for ObjectBox, I also had to add setUp and tearDown methods to remove artefacts in UnitTests: https://github.com/objectbox/objectbox-java/issues/24
2. The library has been added through Cocoapods since is the only PM available (Carthage is experimental for now)
3. I had to make a little change to a test in "Tests/FeedStoreSpecs/XCTestCase+FeedStoreHelpers.swift" because the test was checking the equality between dates but ObjectBox will save/retrieve that rounded. Do you have any tips on how to make the test pass without actually changing it?
4. The following files are generated by the framework, you don't have to review them: "generated/EntityInfo-FeedStoreChallenge.generated.swift", "model-FeedStoreChallenge.json".

Thank you very much! I had a great time with this challenge!


RRReDz:
Closed to investigate the errors related to CI.

PR #122, opened 2021-03-29
ObjectBox implementation - Riccardo Rossi

RRReDz:
Hi Caio & Mike,
Since I've already used Realm and the other solutions (except for Core Data, but I saw that the next lectures you will cover that topic), I tried something new, just to experiment! I hope it's fine for you and will not be too complicated to review, if so, I will re-implement everything with one of the suggested solutions.

Just few info:
1. The in-memory storage is not available for ObjectBox, I also had to add setUp and tearDown methods to remove artefacts in UnitTests: objectbox/objectbox-java#24 (it's for the Java version but still valid for Swift)
2. The library has been added through Cocoapods since is the only PM available (Carthage is experimental for now)
3. I had to make a little change to the CI.yml in order to first run a pod install (in this way you won't find a lot diffs due to the pods) and then clean build and test the workspace instead of the project file. 
4. The following files are generated by the framework, you don't have to review them: "generated/EntityInfo-FeedStoreChallenge.generated.swift", "model-FeedStoreChallenge.json".

I had a great time with this challenge!

Thank you very much!


Tests/FeedStoreSpecs/XCTestCase+FeedStoreHelpers.swift
```diff
@@ -62,7 +62,7 @@ extension XCTestCase {
                 
             case let (.found(expectedFeed, expectedTimestamp), .found(retrievedFeed, retrievedTimestamp)):
                 XCTAssertEqual(retrievedFeed, expectedFeed, file: file, line: line)
-                XCTAssertEqual(retrievedTimestamp, expectedTimestamp, file: file, line: line)
+				XCTAssertEqual(retrievedTimestamp, expectedTimestamp, file: file, line: line)
```
caiozullo, line 65: Undo this change - indentation seems broken.

FeedStoreChallenge/ObjectBoxFeedStore.swift
```diff
@@ -0,0 +1,161 @@
...
+	// MARK: - Private methods
+	
+	private func clearCache() {
+		removeAll(for: Cache.self)
+		removeAll(for: StoreFeed.self)
+	}
+	
+	@discardableResult
+	private func removeAll<T>(for entityType: T.Type = T.self) -> UInt64? where T : ObjectBox.EntityInspectable, T : ObjectBox.__EntityRelatable, T == T.EntityBindingType.EntityType {
+		return try? self.store.box(for: entityType).removeAll()
```
caiozullo, line 114: This operation can fail, but errors are ignored with `try?`.

Errors shouldn't be ignored. 

If there's an error, it should be thrown and the completion closure should be called with the error.

FeedStoreChallenge/ObjectBoxFeedStore.swift
```diff
@@ -0,0 +1,161 @@
...
+	}
+	
+	private func setCache(storeFeeds: [StoreFeed]) {
+		try? self.store.box(for: StoreFeed.self).put(storeFeeds)
```
caiozullo, line 118: This operation can fail, but errors are ignored with `try?`.

Errors shouldn't be ignored. 

If there's an error, it should be thrown and the completion closure should be called with the error.

FeedStoreChallenge/ObjectBoxFeedStore.swift
```diff
@@ -0,0 +1,161 @@
...
+	}
+	
+	private func getCache() -> Cache? {
+		let cacheBox = store.box(for: Cache.self)
+		return try? cacheBox.all().first
```
caiozullo, line 123: This operation can fail, but errors are ignored with `try?`.

Errors shouldn't be ignored. 

If there's an error, it should be thrown and the completion closure should be called with the error.

FeedStoreChallenge/ObjectBoxFeedStore.swift
```diff
@@ -0,0 +1,161 @@
...
+			
+			self.clearCache()
+			
+			completion(nil)
+		}
+	}
+	
+	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+		queue.async(flags: .barrier) { [weak self] in
+			guard let self = self else { return }
```
caiozullo, line 77: `self` doesn't need to be `weakified` in this context. So you don't need to deal with optionals and can remove the `guard` statements too.

FeedStoreChallenge/ObjectBoxFeedStore.swift
```diff
@@ -0,0 +1,161 @@
...
+	
+	public init(storeURL: URL) throws {
+		self.storeURL = storeURL
+		self.store = try Store(directoryPath: storeURL.path)
+		self.queue = DispatchQueue(label: "\(type(of: self))", qos: .userInitiated, attributes: .concurrent)
+	}
+	
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		queue.async(flags: .barrier) { [weak self] in
+			guard let self = self else { return }
```
caiozullo, line 67: `self` doesn't need to be `weakified` in this context. So you don't need to deal with optionals and can remove the `guard` statements too.

FeedStoreChallenge/ObjectBoxFeedStore.swift
```diff
@@ -0,0 +1,161 @@
...
+			self.clearCache()
+			self.setCache(storeFeeds: storeFeeds)
+			
+			completion(nil)
+		}
+	}
+	
+	public func retrieve(completion: @escaping RetrievalCompletion) {
+		queue.async { [weak self] in
+			guard let self = self else { return }
```
caiozullo, line 90: `self` doesn't need to be `weakified` in this context. So you don't need to deal with optionals and can remove the `guard` statements too.

FeedStoreChallenge/ObjectBoxFeedStore.swift
```diff
@@ -0,0 +1,161 @@
...
+		var location: String?
+		var url: String = ""
+		var cache: ToOne<Cache> = nil
+		
+		init(modelId: String = "", description: String? = nil, location: String? = nil, url: String = "", cache: Cache? = nil) {
+			self.modelId = modelId
+			self.description = description
+			self.location = location
+			self.url = url
+			self.cache.target = cache
```
caiozullo, line 32: Ideally, you should define an interface that prevents clients from creating the model with the wrong state. For example, `modelId` shouldn't be *any* `String` in the initializer. It should be a UUID. And the url should be a `URL`.

And they shouldn't have default values.

```suggestion
		init(modelId: UUID, description: String?, location: String?, url: URL, cache: Cache) {
			self.modelId = modelId.uuidString
			self.description = description
			self.location = location
			self.url = url.absoluteString
			self.cache.target = cache
```

FeedStoreChallenge/ObjectBoxFeedStore.swift
```diff
@@ -0,0 +1,161 @@
...
+	}
+	
+	class Cache: Entity {
+		var id: Id = 0
+		var timestamp: Double
+		// objectbox: backlink = "cache"
+		var feed: ToMany<StoreFeed> = nil
+		
+		init(timestamp: Double = 0.0) {
+			self.timestamp = timestamp
```
caiozullo, line 43: Ideally, you should define an interface that prevents clients from creating the model with the wrong state. For example, `timestamp` shouldn't be *any* `Double` in the initializer. It should be a `Date`. 

And it shouldn't have a default value.

```suggestion
		init(timestamp: Date) {
			self.timestamp = timestamp.timeIntervalSinceReferenceDate
```

Tests/FeedStoreChallengeTests.swift
```diff
@@ -107,50 +118,55 @@ class FeedStoreChallengeTests: XCTestCase, FeedStoreSpecs {
...
+		cache.timestamp = Date()
+
+		let storeFeed = StoreFeed()
+		storeFeed.modelId = "Invalid UUID"
+		storeFeed.description = "A description"
+		storeFeed.location = "A location"
+		storeFeed.url = "Invalid URL"
+		storeFeed.cache.target = cache
+
+		try! store.box(for: StoreFeed.self).put(storeFeed)
```
caiozullo, line 151: This method can throw, so you shouldn't force-try with `try!`. 

This way, if there's an error, it'll generate a test failure instead of a crash.

```suggestion
		try store.box(for: StoreFeed.self).put(storeFeed)
```

caiozullo (commented):
Hi @RRReDz,

I never used ObjectBox, so I'll trust the tests on this one!

You almost passed the challenge. The tests are passing, but there are some points I want you to review.

Please, check my comments and let me know once you're done so I can review it again. Thank you 🙂

FeedStoreChallenge/ObjectBoxFeedStore.swift
RRReDz: Yes, but here's the question: Since I couldn't made that method throw an exception (That doesn't throw when you try to perform any kind of operation from outside to the actual DB file, like deleting/overriding/changing permissions to it. I explained this in the commit history.) isn't just better to treat that method as if "doesn't throw", instead of adding a behaviour (complete with an error when get an exception) that could not be verified? (That will also decrease the global code coverage)

FeedStoreChallenge/ObjectBoxFeedStore.swift
RRReDz: Like here: https://github.com/essentialdevelopercom/ios-lead-essentials-feed-store-challenge/pull/122#discussion_r604303177

FeedStoreChallenge/ObjectBoxFeedStore.swift
RRReDz: Same as here: #122 (comment)

FeedStoreChallenge/ObjectBoxFeedStore.swift
RRReDz: Right! Just one consideration: I had to add default parameters values to satisfy the following code generated from a ObjectBox's script `generated/EntityInfo-FeedStoreChallenge.generated.swift`:

`... let entity = StoreFeed() ...`

For this reason i created an initializer how you correctly pointed out and one to satisfy the previous one:

`init() {
	self.modelId = ""
	self.description = nil
	self.location = nil
	self.url = ""
	self.cache = nil
}`

FeedStoreChallenge/ObjectBoxFeedStore.swift
RRReDz: Similar to here: https://github.com/essentialdevelopercom/ios-lead-essentials-feed-store-challenge/pull/122#discussion_r604378777

FeedStoreChallenge/ObjectBoxFeedStore.swift
caiozullo: Hi @RRReDz, it's important to handle errors correctly - even if you can't simulate them during tests.

FeedStoreChallenge/ObjectBoxFeedStore.swift
caiozullo: Hi @RRReDz, I meant it shouldn't have default values for the params in the initializer. But the properties can have default values to satisfy the framework requirements.

caiozullo:
Hi @RRReDz,

- All tests are passing ✅ 

Great job!

RRReDz:
Thank you very much for your time! @caiozullo

PR #131, opened 2021-04-25
Danil Lakhtin - Feed Store Challenge

danillahtin:


FeedStoreChallenge/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,110 @@
...
+        let group = DispatchGroup()
+        group.enter()
+
+        var error: Error?
+        loadPersistentStores(completionHandler: {
+            error = $1.map(Error.loadPersistentStores)
+            group.leave()
+        })
+
+        group.wait()
```
caiozullo, line 99: `loadPersistentStores` already loads synchronously by default (even though it has a completion handler), so the dispatch group is not needed.

To make it load asynchronously, you need to set `shouldAddStoreAsynchronously` in the store description. https://developer.apple.com/documentation/coredata/nspersistentstoredescription/1640599-shouldaddstoreasynchronously

You can remove the dispatch group:

```suggestion
        var error: Error?
        loadPersistentStores(completionHandler: {
            error = $1.map(Error.loadPersistentStores)
        })
```

FeedStoreChallenge/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// MARK: - Helpers
+
+private extension NSManagedObjectContext {
+    func modify(_ block: @escaping (NSManagedObjectContext) throws -> (), completion: @escaping (Error?) -> ()) {
+        perform { [self] in
+            do {
+                try block(self)
+                try self.save()
```
caiozullo, line 55: Some operations may not change the context at all (e.g., trying to delete a cache that is already empty). Calling save wouldn't be needed in those cases, but this method doesn't have knowledge about the operations to decide when to call save or not. So you can check if there are changes in the context before calling save:

```suggestion
                try block(self)
                if self.hasChanges {
                        try self.save()
                }
```

From the docs:

*Always verify that the context has uncommitted changes (using the hasChanges property) before invoking the save: method. Otherwise, Core Data may perform unnecessary work.*

https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506866-save

FeedStoreChallenge/CoreDataFeedStore/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,110 @@
...
+
+    public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+        context.modify(ManagedCache.removeIfExists, completion: completion)
+    }
+
+    public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
+        context.modify({ context in
+            try ManagedCache.removeIfExists(in: context)
+            let cache = ManagedCache(with: feed, timestamp: timestamp, context: context)
+            context.insert(cache)
```
caiozullo, line 28: Calling `context.insert` is not needed since the `ManagedCache` instance was already created using the `context` via `ManagedCache(context: context)`.

You can remove this line:

```suggestion
```

caiozullo (commented):
Hi @danillahtin, you almost passed the challenge!

It's missing the integration tests.

Also, please check my review comments. Let me know once you're done so I can review it again.

Thank you 🙂

danillahtin:
@caiozullo Hey! The challenge is ready

caiozullo (commented):
Hi @danillahtin,

- All tests are passing ✅ 

Great job!

PR #145, opened 2021-05-17
Alex Kostenko - Feed Store Challenge

oligazar:


FeedStoreChallenge/CoreDataFeedStore.xcdatamodeld/CoreDataFeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="17709" systemVersion="19H1030" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <elements/>
+</model>
```
caiozullo, line 4: This file is not used. Please, remove it.

FeedStoreChallenge/Infrastructure/CoreData/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,54 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				try ManagedCache.find(in: context).map(context.delete)
+				if (context.hasChanges) {
```
caiozullo, line 65: The parenthesis can be removed:

```suggestion
				if context.hasChanges {
```

FeedStoreChallenge/Infrastructure/CoreData/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,54 @@ public final class CoreDataFeedStore: FeedStore {
...
+					try context.save()
+				}
```
caiozullo, line 67: Another way to ensure the context is only saved when there are changes is to use `.map` instead of an `if` statement:

```suggestion
				try ManagedCache.find(in: context)
					.map(context.delete)
					.map(context.save)
```

FeedStoreChallenge/Infrastructure/CoreData/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="17709" systemVersion="19H1030" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="ManagedCache" representedClassName="ManagedCache" syncable="YES">
+        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feed" optional="YES" toMany="YES" deletionRule="Nullify" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
```
caiozullo, line 5: The `feed` relationship `deletionRule` should be `Cascade` to prevent orphaned objects in the DB (when we delete the `ManagedCache` instance, we want all images in the `feed` property to be deleted as well!).

```suggestion
        <relationship name="feed" optional="YES" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
```

FeedStoreChallenge/Infrastructure/CoreData/ManagedCache.swift
```diff
@@ -0,0 +1,35 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Alexander on 17/5/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedCache)
+internal class ManagedCache: NSManagedObject {
```
caiozullo, line 12: In Swift, `internal` is the default access level so it doesn't need to be set explicitly.

`internal` can be removed:

```suggestion 
class ManagedCache: NSManagedObject {
```

---

Please, make sure to remove it everywhere to make the code consistent.

FeedStoreChallenge/Infrastructure/CoreData/ManagedCache.swift
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	internal var localFeed: [LocalFeedImage] {
+		return feed.compactMap { ($0 as? ManagedFeedImage)?.local }
+	}
+
+	internal static func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		guard let entityName = ManagedCache.entity().name else { return nil }
+		let request = NSFetchRequest<ManagedCache>(entityName: entityName)
+		request.returnsDistinctResults = false
```
caiozullo, line 31: `returnsDistinctResults` is not needed in this context.

From the docs:

"This value is used only if a value has been set for propertiesToFetch." - https://developer.apple.com/documentation/coredata/nsfetchrequest/1506344-returnsdistinctresults

You can remove this line:

```suggestion
```

caiozullo (commented):
Hi @oligazar, you almost passed the challenge!

Thank you 🙂

FeedStoreChallenge/Infrastructure/CoreData/ManagedCache.swift
oligazar: Hi, @caiozullo, thanks for your review and suggestions! 

I applied changes from Xcode (but I think doing it using github's UI would be easier and more representative, as after pushing the changes, your suggestions don't correlate anymore with the updated code. Which way of applying suggestions is preferred in your opinion?). 

Regarding `CoreDataFeedStore`. I deleted it from Xcode, but for some reason it wasn't deleted from the file system and this change wasn't represented in git. I tried restoring the file and deleting it again, but this change still wasn't tracked in git. So I ended up deleting the file from github itself. What exactly can I do to avoid this next time?

Thanks for your time and the great course!

FeedStoreChallenge/Infrastructure/CoreData/ManagedCache.swift
```diff
@@ -0,0 +1,34 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Alexander on 17/5/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedCache)
+class ManagedCache: NSManagedObject {
```
caiozullo, line 12: This class is not meant to be subclassed so you can mark it as final:

```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/Infrastructure/CoreData/ManagedFeedImage.swift
```diff
@@ -0,0 +1,35 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Alexander on 17/5/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedFeedImage)
+class ManagedFeedImage: NSManagedObject {
```
caiozullo, line 12: This class is not meant to be subclassed so you can mark it as final:

```suggestion
final class ManagedFeedImage: NSManagedObject {
```

FeedStoreChallenge/Infrastructure/CoreData/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,54 @@ public final class CoreDataFeedStore: FeedStore {
 	}
 
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
```
caiozullo, line 32: Another way of capturing the `context` inside the closure without capturing `self` is to add it to the capture list as `[context] in`:

```suggestion
		context.perform { [context] in
```

If you follow this suggestion everywhere in the code, you can delete the `private func perform(completion:)` method ✅

caiozullo (commented):
Hi @oligazar, 

- All tests are passing ✅ 

Great job!

---

Regarding your questions:

> 1. I applied changes from Xcode (but I think doing it using github's UI would be easier and more representative, as after pushing the changes, your suggestions don't correlate anymore with the updated code. Which way of applying suggestions is preferred in your opinion?). 

I recommend you do it on Xcode because some changes may affect other parts of the code that GitHub wouldn't track and can lead to failing tests or build errors.

---

> 2. Regarding `CoreDataFeedStore`. I deleted it from Xcode, but for some reason it wasn't deleted from the file system and this change wasn't represented in git. I tried restoring the file and deleting it again, but this change still wasn't tracked in git. So I ended up deleting the file from github itself. What exactly can I do to avoid this next time?

When deleting a file from Xcode, make sure to select 'Move to trash' to remove it from the file system. Otherwise, Xcode removes only its reference from the project (but keeps the file in the file system). Another way is to manually delete the file from the file system.

Keep up the great work! 🙂

PR #146, opened 2021-05-18
extract functions into extensions

adiallo22:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,51 @@ public final class CoreDataFeedStore: FeedStore {
...
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		perform { (context) in
+			do {
+				if let cache = try ManagedCache.find(in: context) {
+					context.delete(cache)
+				}
+				try context.save()
```
caiozullo, line 66: `try context.save()` should be inside the if statement block because the context should only be saved if there are changes to it. Otherwise, Core Data may perform unnecessary work.

 ```suggestion
				if let cache = try ManagedCache.find(in: context) {
					context.delete(cache)
					try context.save()
				}
```

caiozullo (commented):
Hi @adiallo22,

- All tests are passing ✅ 

I left only one comment. Feel free to push changes.

Great job!

FeedStoreChallenge/CoreDataFeedStore.swift
adiallo22: Thank you guys for providing is much value 
I'm learning a lot

caiozullo:
Nice one!

PR #149, opened 2021-05-29
Darren Findlay - Feed Store Challenge.

darrenfindlay:


FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,62 @@
...
+	@NSManaged var feed: NSOrderedSet
+
+	internal static func newSingleCache(in context: NSManagedObjectContext) throws -> ManagedCache {
+		try findFirst(in: context).map(context.delete)
+		return ManagedCache(context: context)
+	}
+
+	static func findFirst(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		let request = NSFetchRequest<ManagedCache>(entityName: ManagedCache.entity().name!)
+		return try context.fetch(request).first
```
danillahtin, line 24: To slightly increase CoreData performance `request.returnsObjectsAsFaults` can be set to `false`.

```suggestion
		request.returnsObjectsAsFaults = false
		return try context.fetch(request).first
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	var localFeed: [LocalFeedImage] {
+		feed.map { $0 as! ManagedFeedImage }
```
danillahtin, line 28: Instead of `map` with force cast, it could be `compactMap` with optional casting

```suggestion
		feed.compactMap { $0 as? ManagedFeedImage }
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,62 @@
...
+	static func from(feed: [LocalFeedImage], timestamp: Date, in context: NSManagedObjectContext) -> ManagedCache {
+		let cache = ManagedCache(context: context)
+		cache.timestamp = timestamp
+		cache.feed = ManagedFeedImage.images(from: feed, in: context)
+		return cache
+	}
+}
+
+@objc(ManagedFeedImage)
+class ManagedFeedImage: NSManagedObject {
```
danillahtin, line 41: This class is not intended to be inherited thus it should be `final`, please make it `final`

```suggestion
final class ManagedFeedImage: NSManagedObject {
```

danillahtin (commented):
Hi @darrenfindlay, you almost passed the challenge!

Thank you 🙂

darrenfindlay:
Hi @danillahtin , Thanks for the feedback! I've implemented all changes you suggested.

danillahtin (commented):
Hi @darrenfindlay,

- All tests are passing ✅ 

Great job!

PR #150, opened 2021-05-30
Chris Wong - Feed Store Challenge

therealchriswong:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,47 @@ public final class CoreDataFeedStore: FeedStore {
 	}
 
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		fatalError("Must be implemented")
+		let context = self.context
+		context.perform {
```
danillahtin, line 33: During the course this way of having a local reference to the context to avoid capturing self inside of the `context.perform` block is taught.
The same can be achieved using capture list.

```suggestion
		context.perform { [context] in
```

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="17709" systemVersion="20E241" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="ManagedCache" representedClassName="ManagedCache" syncable="YES">
+        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feed" optional="YES" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
```
danillahtin, line 5: `feed` property is required, please make it non-optional

```suggestion
        <relationship name="feed" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
```

FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,32 @@
...
+//
+import CoreData
+
+@objc(ManagedFeedImage)
+final class ManagedFeedImage: NSManagedObject {
+	@NSManaged var id: UUID
+	@NSManaged var imageDescription: String?
+	@NSManaged var location: String?
+	@NSManaged var url: URL
+	@NSManaged var cache: ManagedCache
```
danillahtin, line 16: `cache` property is marked as optional in `FeedStore.xcdatamodel`, please make it optional to make the model consistent.

```suggestion
	@NSManaged var cache: ManagedCache?
```

danillahtin (commented):
Hi @therealchriswong, you almost passed the challenge!

Thank you 🙂

FeedStoreChallenge/CoreDataFeedStore.swift
therealchriswong: Would this be an alternative way to write it?


```
context.perform {
   [weak self] in
  guard let context = self?.context else { return }
  ....
```

therealchriswong:
@danillahtin  Updates are done. Please review. Thanks!

FeedStoreChallenge/CoreDataFeedStore.swift
danillahtin: It is not 100% the same. In your example commands after `guard ... let` will not be executed in case `self` is deallocated whereas in `context.perform { [context] in ... }` they will as it has no check for `self` deallocation

danillahtin:
Hi @therealchriswong,

The CI build failed.

Please, fix all build failures and let me know once you're done so I can start the review. Thank you 🙂

FeedStoreChallenge/CoreDataFeedStore.swift
therealchriswong: You're right. It's a using strong reference to avoid a weak self.  I changed it to { [context] in ... }  as suggested.  @danillahtin

therealchriswong:
@danillahtin It's not failing for me locally. I only changed the capture list.

danillahtin:
@therealchriswong Let me check

danillahtin (commented):
Hi @therealchriswong,

- All tests are passing ✅ 

Great job!

PR #152, opened 2021-06-02
Naveen Tg - Feed Store Challenge

NaveenTg:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,48 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		performContext { context in
+			do {
+				try ManagedCache.fetch(in: context).map(context.delete)
+				try context.save()
```
danillahtin, line 64: If there's no cache then `context.delete` will not delete anything and `context` will have no changes.
`try context.save()` should be called only if there're some changes,  otherwise, Core Data may perform unnecessary work.

```suggestion
				try ManagedCache.fetch(in: context).map(context.delete).map(context.save)
```

`Optional.map` only executes the closure if there's a value. So if `ManagedCache.fetch` returns `nil`, it won't execute the `context.delete` or `context.save` ✅

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,48 @@ public final class CoreDataFeedStore: FeedStore {
...
+				completion(nil)
+			} catch {
+				context.rollback()
+				completion(error)
+			}
+		}
+	}
+
+	private func performContext(_ action: @escaping (NSManagedObjectContext) -> Void) {
+		context.perform { [self] in action(context) }
```
danillahtin, line 74: There's no need to capture `self` because it can cause `CoreDataFeedStore` not being deallocated in some cases.
It can be optimised just to capturing `context` only using capture list.

```suggestion
		context.perform { [context] in action(context) }
```

danillahtin (commented):
Hi @NaveenTg, you almost passed the challenge!

Thank you 🙂

NaveenTg:
@danillahtin ,
Thanks for the input , i have made the changes. Can you review the same.

danillahtin (commented):
Hi @NaveenTg,

- All tests are passing ✅ 

Great job!

PR #153, opened 2021-06-05
Fei Wen - Feed Store Challenge

fnytrip:


FeedStoreChallenge/CoreDataModel/ManagedCache+CoreDataClass.swift
```diff
@@ -0,0 +1,30 @@
...
+//  Created by Fei Wen on 6/5/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+//
+
+import Foundation
+import CoreData
+
+@objc(ManagedCache)
+public class ManagedCache: NSManagedObject {
```
thiagopenna, line 14: By default Xcode generates two files for CoreData classes. It is ok to move all their content into one file. Having a 2 line declaration in a separate file with extensions in another one doesn't bring any value.

Please move extensions from `ManagedCache+CoreDataProperties.swift` here.

FeedStoreChallenge/CoreDataModel/ManagedFeedImage+CoreDataClass.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Created by Fei Wen on 6/5/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+//
+
+import Foundation
+import CoreData
+
+@objc(ManagedFeedImage)
+public class ManagedFeedImage: NSManagedObject {
```
thiagopenna, line 14: `ManagedFeedImage` is an implementation detail of `CoreDataFeedStore` thus it should not be public.
Also it is not intended to be inherited from, thus it should be final.

Please remove `public` modifier and make the class `final`.
```suggestion
final class ManagedFeedImage: NSManagedObject {
```

FeedStoreChallenge/CoreDataModel/ManagedFeedImage+CoreDataProperties.swift
```diff
@@ -0,0 +1,25 @@
...
+extension ManagedFeedImage {
+	@nonobjc public class func fetchRequest() -> NSFetchRequest<ManagedFeedImage> {
+		return NSFetchRequest<ManagedFeedImage>(entityName: "ManagedFeedImage")
+	}
+
+	@NSManaged public var id: UUID
+	@NSManaged public var imageDescription: String?
+	@NSManaged public var location: String?
+	@NSManaged public var url: URL
+	@NSManaged public var cache: ManagedCache
```
thiagopenna, line 22: `ManagedFeedImage` is an implementation detail of `CoreDataFeedStore` thus its properties should not be public.
```suggestion
	@NSManaged var id: UUID
	@NSManaged var imageDescription: String?
	@NSManaged var location: String?
	@NSManaged var url: URL
	@NSManaged var cache: ManagedCache
```

FeedStoreChallenge/CoreDataModel/ManagedFeedImage+CoreDataProperties.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+//
+
+import Foundation
+import CoreData
+
+extension ManagedFeedImage {
+	@nonobjc public class func fetchRequest() -> NSFetchRequest<ManagedFeedImage> {
+		return NSFetchRequest<ManagedFeedImage>(entityName: "ManagedFeedImage")
+	}
```
thiagopenna, line 16: This method is not used, please remove it
```suggestion
```

FeedStoreChallenge/CoreDataModel/ManagedFeedImage+CoreDataProperties.swift
```diff
@@ -0,0 +1,25 @@
...
+
+	@NSManaged public var id: UUID
+	@NSManaged public var imageDescription: String?
+	@NSManaged public var location: String?
+	@NSManaged public var url: URL
+	@NSManaged public var cache: ManagedCache
+}
+
+extension ManagedFeedImage: Identifiable {}
```
thiagopenna, line 25: This extension is not used, please remove it
```suggestion
```

FeedStoreChallenge/CoreDataModel/ManagedFeedImage+CoreDataClass.swift
thiagopenna, line 14: By default Xcode generates two files for CoreData classes. It is ok to move all their content into one file. Having a 2 line declaration in a separate file with extensions in another one doesn't bring any value.

Please move extensions from `ManagedFeedImage+CoreDataProperties.swift` here.

thiagopenna (commented):
Hi @fnytrip, you almost passed the challenge!

Thank you 🙂

FeedStoreChallenge/CoreDataModel/ManagedCache+CoreDataClass.swift
```diff
@@ -0,0 +1,30 @@
...
+	static func findFirst(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		let request = ManagedCache.fetchRequest() as NSFetchRequest<ManagedCache>
+		request.returnsObjectsAsFaults = false
+		return try context.fetch(request).first
+	}
+
+	static func deleteAll(in context: NSManagedObjectContext) throws {
+		let request = ManagedCache.fetchRequest() as NSFetchRequest<ManagedCache>
+		request.returnsDistinctResults = false
+		let _ = try context.fetch(request).map { context.delete($0) }
```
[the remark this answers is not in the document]
fnytrip: Actually I don't like that we are relying on an assumption that there is always only one instance of ManagedCache, that's why I name my func as `deleteAll` and not reusing the same code in `findFirst`. 

But anyway, I changed it per your suggestion.

FeedStoreChallenge/CoreDataModel/ManagedCache+CoreDataClass.swift
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	var localFeeds: [LocalFeedImage] {
```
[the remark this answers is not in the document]
fnytrip: Good catch, updated.

fnytrip:
@thiagopenna Thank you for reviewing. I've updated the PR.

FeedStoreChallenge/CoreDataModel/ManagedCache+CoreDataClass.swift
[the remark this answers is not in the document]
thiagopenna: Sorry, you're absolutely right, I didn't pay enough attention to which class guaranteed the unique behavior, and assumed it was the `ManagedCache`. Although there's some duplication on creating the `FetchRequest` and fetching results, it would not be correct to change the implemented logic in this refactor. My suggestion then is to extract the duplicated portion of code to a helper function, something like:
```swift
	static func findFirst(in context: NSManagedObjectContext) throws -> ManagedCache? {
		return try findAll(in: context).first
	}

	static func deleteAll(in context: NSManagedObjectContext) throws {
		try findAll(in: context).forEach(context.delete)
	}

	private static func findAll(in context: NSManagedObjectContext) throws -> [ManagedCache] {
		let request = ManagedCache.fetchRequest() as NSFetchRequest<ManagedCache>
		return try context.fetch(request)
	}
```

thiagopenna (commented):
Hi @fnytrip,

- All tests are passing ✅ 

I left just one review comment. Feel free to push changes.

Great job!

FeedStoreChallenge/CoreDataModel/ManagedCache+CoreDataClass.swift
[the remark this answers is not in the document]
fnytrip: @thiagopenna Thank you. I pushed the last commit with your suggested change. Cheers!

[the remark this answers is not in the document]
thiagopenna: Well done!

PR #155, opened 2021-06-11
Chad Galloway - Feed Store Challenge

cwgallo:


FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
...
+    <entity name="ManagedCache" representedClassName="ManagedCache" syncable="YES">
+        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feed" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
+    </entity>
+    <entity name="ManagedFeedImage" representedClassName="ManagedFeedImage" syncable="YES">
+        <attribute name="id" attributeType="UUID" usesScalarValueType="NO"/>
+        <attribute name="imageDescription" optional="YES" attributeType="String"/>
+        <attribute name="location" optional="YES" attributeType="String"/>
+        <attribute name="url" attributeType="URI"/>
+        <relationship name="cache" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="ManagedCache" inverseName="feed" inverseEntity="ManagedCache"/>
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,50 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				try ManagedCache.find(in: context).map(context.delete)
+				try context.save()
```
thiagopenna, line 65: In the case where there was no previously stored cache, `context.delete` will not be called, thus there will be no changes in the context and no need to save it. Otherwise, Core Data may perform unnecessary work.
It's possible to use `context.hasChanges` to check if saving is needed, or use `.map` instead:
```suggestion
				try ManagedCache.find(in: context).map(context.delete).map(context.save)
```

thiagopenna (commented):
Hi @cwgallo,

- All tests are passing ✅ 

Great job!

PR #157, opened 2021-06-12
Sam Yang - Feed Store Challenge

startover205:


FeedStoreChallenge/Infrastructure/Core Data/ManagedCache.swift
```diff
@@ -0,0 +1,25 @@
...
+	@NSManaged internal var timestamp: Date
+	@NSManaged internal var feed: NSOrderedSet
+
+	internal var localFeed: [LocalFeedImage] {
+		return self.feed.compactMap { ($0 as? ManagedFeedImage)?.local }
+	}
+
+	internal static func find() throws -> ManagedCache? {
+		try ManagedCache.fetchRequest().execute().first as? ManagedCache
+	}
```
thiagopenna, line 24: By calling `execute()` without injecting a context we depend on `CoreData`'s implementation to get the current `context` based on which `queue` called the execute, and the way the queue is created might change in the future. To avoid this problem, we can inject the context on which to perform the fetch request.
```swift
static func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
	return try context.fetch(ManagedCache.fetchRequest()).first as? ManagedCache
}
```
Also, to slightly increase CoreData performance `request.returnsObjectsAsFaults` can be set to `false`.

```suggestion
	static func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
		let request = ManagedCache.fetchRequest()
		request.returnsObjectsAsFaults = false
		return try context.fetch(request).first as? ManagedCache
	}
```

FeedStoreChallenge/Infrastructure/Core Data/ManagedCache.swift
```diff
@@ -0,0 +1,25 @@
...
+import CoreData
+
+@objc(ManagedCache)
+internal final class ManagedCache: NSManagedObject {
+	@NSManaged internal var timestamp: Date
+	@NSManaged internal var feed: NSOrderedSet
+
+	internal var localFeed: [LocalFeedImage] {
+		return self.feed.compactMap { ($0 as? ManagedFeedImage)?.local }
+	}
```
thiagopenna, line 20: The convention in this codebase is to omit the default access-level of `internal`, so we should omit them.
```suggestion
final class ManagedCache: NSManagedObject {
	@NSManaged var timestamp: Date
	@NSManaged var feed: NSOrderedSet

	var localFeed: [LocalFeedImage] {
		return self.feed.compactMap { ($0 as? ManagedFeedImage)?.local }
	}
```

FeedStoreChallenge/Infrastructure/Core Data/ManagedFeedImage.swift
```diff
@@ -0,0 +1,34 @@
...
+import CoreData
+
+@objc(ManagedFeedImage)
+internal final class ManagedFeedImage: NSManagedObject {
+	@NSManaged internal var id: UUID
+	@NSManaged internal var imageDescription: String?
+	@NSManaged internal var location: String?
+	@NSManaged internal var url: URL
+
+	internal static func images(from feed: [LocalFeedImage], in context: NSManagedObjectContext) -> NSOrderedSet {
```
thiagopenna, line 20: The convention in this codebase is to omit the default access-level of internal, so we should omit them.
```suggestion
final class ManagedFeedImage: NSManagedObject {
	@NSManaged var id: UUID
	@NSManaged var imageDescription: String?
	@NSManaged var location: String?
	@NSManaged var url: URL

	static func images(from feed: [LocalFeedImage], in context: NSManagedObjectContext) -> NSOrderedSet {
```

FeedStoreChallenge/Infrastructure/Core Data/ManagedFeedImage.swift
```diff
@@ -0,0 +1,34 @@
...
+			let image = ManagedFeedImage(context: context)
+			image.id = $0.id
+			image.imageDescription = $0.description
+			image.location = $0.location
+			image.url = $0.url
+			return image
+		})
+	}
+
+	internal var local: LocalFeedImage {
```
thiagopenna, line 31: The convention in this codebase is to omit the default access-level of internal, so we should omit them.
```suggestion
	var local: LocalFeedImage {
```

FeedStoreChallenge/Infrastructure/Core Data/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
...
+    <entity name="ManagedCache" representedClassName="ManagedCache" syncable="YES">
+        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feed" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
+    </entity>
+    <entity name="ManagedFeedImage" representedClassName="ManagedFeedImage" syncable="YES">
+        <attribute name="id" attributeType="UUID" usesScalarValueType="NO"/>
+        <attribute name="imageDescription" optional="YES" attributeType="String"/>
+        <attribute name="location" optional="YES" attributeType="String"/>
+        <attribute name="url" attributeType="URI"/>
+        <relationship name="cache" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="ManagedCache" inverseName="feed" inverseEntity="ManagedCache"/>
```
thiagopenna, line 12: The `FeedStore.xcdatamodeld` declares a `ManagedFeedImage`'s `cache` relationship as `optional`, but the `ManagedFeedImage` implementation does not have a `cache` property. This could lead to runtime issues since having a `ManagedFeedImage` without a related cache is a representable state.
Both the `DataModel` and the actual class implementation should represent the same entity, please add a`@NSManaged var cache` to the class, also respecting optionality.

FeedStoreChallenge/Infrastructure/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,86 @@
...
+				completion(error)
+			}
+		}
+	}
+
+	private static func deleteCacheFeedIfNeeded(context: NSManagedObjectContext) throws {
+		if let cache = try ManagedCache.find() {
+			context.delete(cache)
+		}
+	}
```
thiagopenna, line 80: This method can cause confusion, since it reads as `CoreDataFeedStore.deleteCacheFeedIfNeeded`, but does not save the deletion nor calls inside a `context.perform` block. It would make sense to move it to `ManagedCache` as a static method, similar to `find()`: with no saves to context.
Also, I believe there's a typo in this method's name: it should read `deleteCachedFeedIfNeeded`.

FeedStoreChallenge/Infrastructure/Core Data/CoreDataFeedStore.swift
```diff
@@ -0,0 +1,86 @@
...
+				completion(error)
+			}
+		}
+	}
+
+	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
+		perform { context in
+			do {
+				try CoreDataFeedStore.deleteCacheFeedIfNeeded(context: context)
+				try context.save()
```
thiagopenna, line 67: In the case where there was no previously stored cache, `context.delete` will not be called, thus there will be no changes in the context and no need to save it. Otherwise, Core Data may perform unnecessary work.
It's possible to use `context.hasChanges` to check if saving is needed.
```suggestion
				if context.hasChanges {
					try context.save()
				}
```

thiagopenna (commented):
Hi @startover205, you almost passed the challenge!

Thank you 🙂

startover205:
Hi @thiagopenna, thank you for the review!😃
I've applied the changes. 
Hope you can take a look. Thank you!

thiagopenna:
Hi @startover205,

- All tests are passing ✅ 

Great job!

PR #158, opened 2021-06-13
Andreas Link - Feed Store Challenge

LinkAndreas:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,80 @@ public final class CoreDataFeedStore: FeedStore {
...
+				completion(error)
+			}
+		}
+	}
+}
+
+private extension NSOrderedSet {
+	func toLocal() -> [LocalFeedImage] {
+		array
+			.compactMap { $0 as? ManagedFeedImage }
```
thiagopenna, line 83: Since `NSOrderedSet` conforms to `Sequence`, it is possible to call `compactMap` directly, there's no need to convert it to an `array`. 
```suggestion
		compactMap { $0 as? ManagedFeedImage }
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,80 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
-		fatalError("Must be implemented")
+		context.perform { [context] in
+			do {
+				let existingCache = try ManagedCache.fetch(in: context)
+				existingCache.map(context.delete)
```
thiagopenna, line 49: This logic is duplicated between methods `insert` and `delete` (lines 67 and 68). Also, since `fetch` returns an optional `ManagedCache`, the name `existingCache` is misleading since it can be `nil`. 
We can extract this logic to a helper method, or inline it as a single operation making use of optional `map`:
```suggestion
				try ManagedCache.fetch(in: context).map(context.delete)
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,80 @@ public final class CoreDataFeedStore: FeedStore {
...
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		context.perform { [context] in
+			do {
+				let existingCache = try ManagedCache.fetch(in: context)
+				existingCache.map(context.delete)
+
+				try context.save()
```
thiagopenna, line 70: The same problem from my previous comment applies here. Also, in the case where there was no previously stored cache, `context.delete` will not be called, thus there will be no changes in the context and no need to save it. Otherwise, Core Data may perform unnecessary work.
It's possible to use `context.hasChanges` to check if saving is needed, or use another `.map` instead:
```suggestion
				try ManagedCache.fetch(in: context).map(context.delete).map(context.save)
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,80 @@ public final class CoreDataFeedStore: FeedStore {
...
+					description: managedFeedImage.imageDescription,
+					location: managedFeedImage.location,
+					url: managedFeedImage.url
+				)
+			}
+	}
+}
+
+private extension Array where Element == LocalFeedImage {
+	func toManaged(context: NSManagedObjectContext) -> NSOrderedSet {
```
thiagopenna, line 96: This function name can be confusing, since it's being called with `feed.toManaged(context:`, while actually returning a `NSOrderedSet` of `ManagedFeedImage`s.
A suggestion would be to move this logic as a static method in `ManagedFeedImage`, and it would read as:
```swift
ManagedFeedImage.images(from: feed, in: context)
```
Another option would be renaming it so its called as:
```swift
feed.managedImages(in: context)
```
These are suggestions to improve this function naming and clarify what type of object it returns.
```suggestion
	func managedImages(in context: NSManagedObjectContext) -> NSOrderedSet {
```

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
...
+    <entity name="ManagedCache" representedClassName="ManagedCache" syncable="YES">
+        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feed" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
+    </entity>
+    <entity name="ManagedFeedImage" representedClassName="ManagedFeedImage" syncable="YES">
+        <attribute name="id" attributeType="UUID" usesScalarValueType="NO"/>
+        <attribute name="imageDescription" optional="YES" attributeType="String"/>
+        <attribute name="location" optional="YES" attributeType="String"/>
+        <attribute name="url" attributeType="URI"/>
+        <relationship name="cache" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="ManagedCache" inverseName="feed" inverseEntity="ManagedCache"/>
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,22 @@
...
+//
+//
+
+import Foundation
+import CoreData
+
+@objc(ManagedCache)
+internal class ManagedCache: NSManagedObject {
+	@NSManaged internal var timestamp: Date
+	@NSManaged internal var feed: NSOrderedSet
```
thiagopenna, line 16: The convention in this codebase is to omit the default access level keyword of `internal`, please remove it.
```suggestion
	@NSManaged var timestamp: Date
	@NSManaged var feed: NSOrderedSet
```

FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,20 @@
...
+import Foundation
+import CoreData
+
+@objc(ManagedFeedImage)
+internal class ManagedFeedImage: NSManagedObject {
+	@NSManaged internal var id: UUID
+	@NSManaged internal var imageDescription: String?
+	@NSManaged internal var location: String?
+	@NSManaged internal var url: URL
+	@NSManaged internal var cache: ManagedCache
```
thiagopenna, line 19: As per my previous comment, the `internal` keyword should be omitted.
```suggestion
	@NSManaged var id: UUID
	@NSManaged var imageDescription: String?
	@NSManaged var location: String?
	@NSManaged var url: URL
	@NSManaged var cache: ManagedCache
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,22 @@
...
+//  Created by Andreas Link on 13.06.21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+//
+
+import Foundation
+import CoreData
+
+@objc(ManagedCache)
+internal class ManagedCache: NSManagedObject {
```
thiagopenna, line 14: As per my previous comments, the `internal` keyword should be omitted.
Also, this class is not supposed to be subclassed, so it should be marked as final.
```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,22 @@
...
+	@NSManaged internal var timestamp: Date
+	@NSManaged internal var feed: NSOrderedSet
+
+	static func fetch(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		let request = NSFetchRequest<ManagedCache>(entityName: String(describing: ManagedCache.self))
+		return try context.fetch(request).first
```
thiagopenna, line 20: To slightly increase CoreData performance `request.returnsObjectsAsFaults` can be set to `false`.
```suggestion
		let request = NSFetchRequest<ManagedCache>(entityName: String(describing: ManagedCache.self))
		request.returnsObjectsAsFaults = false
		return try context.fetch(request).first
```

thiagopenna (commented):
Hi @LinkAndreas ,

- All tests are passing ✅ 

Great job!

PR #159, opened 2021-06-14
Istvan Balogh - Feed Store Challenge

steven851007:
I've finished the feed store challenge, thank you for the review!

### DIFF
--- a/FeedStoreChallenge/CoreDataFeedStore.swift
+++ b/FeedStoreChallenge/CoreDataFeedStore.swift
@@ -29,7 +29,7 @@ public final class CoreDataFeedStore: FeedStore {
 	}
 
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		fatalError("Must be implemented")
+		completion(.empty)
 	}
 
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
--- a/FeedStoreChallenge/CoreDataFeedStore.swift
+++ b/FeedStoreChallenge/CoreDataFeedStore.swift
@@ -76,7 +76,7 @@ public final class CoreDataFeedStore: FeedStore {
 				}
 				completion(nil)
 			} catch {
-				
+				completion(error)
 			}
 		}
 	}
--- a/FeedStoreChallenge/CoreDataFeedStore.swift
+++ b/FeedStoreChallenge/CoreDataFeedStore.swift
@@ -11,6 +11,7 @@ public final class CoreDataFeedStore: FeedStore {
 	private let container: NSPersistentContainer
 	private let context: NSManagedObjectContext
 	private let cacheStore: ManagedCacheStore
+	private let feedImageStore: ManagedFeedImageStore
 
 	struct ModelNotFound: Error {
 		let modelName: String
@@ -28,6 +29,7 @@ public final class CoreDataFeedStore: FeedStore {
 		)
 		context = container.newBackgroundContext()
 		cacheStore = ManagedCacheStore(context: context)
+		feedImageStore = ManagedFeedImageStore(context: context)
 	}
 
 	public func retrieve(completion: @escaping RetrievalCompletion) {
@@ -49,12 +51,12 @@ public final class CoreDataFeedStore: FeedStore {
 	}
 
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
-		let context = context
 		let cacheStore = cacheStore
+		let feedImageStore = feedImageStore
 		context.perform {
 			do {
 				let cache = try cacheStore.existingCacheOrNewOne()
-				let managedFeedImages = Self.createManagedFeedImage(from: feed, insertInto: context)
+				let managedFeedImages = feed.map { feedImageStore.feedImage(from: $0) }
 				cache.timestamp = timestamp
 				cache.feedImages = managedFeedImages
 				completion(nil)
@@ -66,17 +68,6 @@ public final class CoreDataFeedStore: FeedStore {
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
 		fatalError("Must be implemented")
-	}
-
-	private static func createManagedFeedImage(from feed: [LocalFeedImage], insertInto context: NSManagedObjectContext) -> [ManagedFeedImage] {
-		feed.map {
-			let managedFeedImage = ManagedFeedImage(context: context)
-			managedFeedImage.id = $0.id
-			managedFeedImage.descriptionString = $0.description
-			managedFeedImage.location = $0.location
-			managedFeedImage.url = $0.url
-			return managedFeedImage
-		}
 	}
 }
 
--- a/FeedStoreChallenge/CoreDataFeedStore.swift
+++ b/FeedStoreChallenge/CoreDataFeedStore.swift
@@ -54,9 +54,8 @@ public final class CoreDataFeedStore: FeedStore {
 		context.perform {
 			do {
 				let cache = try cacheStore.existingCacheOrNewOne()
-				let managedFeedImages = feed.map { feedImageStore.feedImage(from: $0) }
+				cache.feedImages = feed.map { feedImageStore.feedImage(from: $0) }
 				cache.timestamp = timestamp
-				cache.feedImages = managedFeedImages
 				try cacheStore.save()
 				completion(nil)
 			} catch {
--- a/FeedStoreChallenge/CoreDataFeedStore.swift
+++ b/FeedStoreChallenge/CoreDataFeedStore.swift
@@ -76,6 +76,7 @@ public final class CoreDataFeedStore: FeedStore {
 				}
 				completion(nil)
 			} catch {
+				cacheStore.rollback()
 				completion(error)
 			}
 		}
--- a/FeedStoreChallenge/CoreDataFeedStore.swift
+++ b/FeedStoreChallenge/CoreDataFeedStore.swift
@@ -33,12 +33,10 @@ public final class CoreDataFeedStore: FeedStore {
 	}
 
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		let context = context
+		let cacheStore = cacheStore
 		context.perform {
-			let fetchRequest: NSFetchRequest<ManagedCache> = ManagedCache.fetchRequest()
 			do {
-				let caches = try context.fetch(fetchRequest)
-				if let cache = caches.first {
+				if let cache = try cacheStore.cache() {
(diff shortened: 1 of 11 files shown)


FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="18154" systemVersion="20F71" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="ManagedCache" representedClassName="ManagedCache" syncable="YES">
+        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feedImagesOrderedSet" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage"/>
```
steven851007, line 5: Should we not have tests also for testing if deleting the cache also deletes the `ManagedFeedItem` objects as well from the database? Not setting this to `cascade` will leave the objects in the database, right? If we should test it where that test should be and how it would look like?

FeedStoreChallenge/EntityStores/ManagedCacheStore.swift
```diff
@@ -0,0 +1,52 @@
...
+
+class ManagedCacheStore {
+	private let context: NSManagedObjectContext
+
+	init(context: NSManagedObjectContext) {
+		self.context = context
+	}
+
+	func cache() throws -> ManagedCache? {
+		let newFetchRequest = fetchRequest()
```
steven851007, line 20: It seems the CI still running on Xcode 12.4 as the line `let fetchRequest = fetchRequest()` is failing there but it compiles on Xcode 12.5

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -26,17 +28,63 @@ public final class CoreDataFeedStore: FeedStore {
...
+		feedImageStore = ManagedFeedImageStore(context: context)
 	}
 
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		fatalError("Must be implemented")
+		context.perform { [cacheStore] in
+			do {
+				if let cache = try cacheStore.cache() {
+					let localFeedImages = cache.feedImages.map { $0.localFeedImage }
+					completion(.found(feed: localFeedImages, timestamp: cache.timestamp))
```
thiagopenna, line 40: It's possible to extract this mapping logic to one of the helpers, for example following the same approach of `ManagedFeedImage`'s extension with a `localFeedImage`, it's possible to implement a `ManagedCache` `localFeed` computed property, that returns `[LocalFeedImage]`:
```suggestion
					completion(.found(feed: cache.localFeed, timestamp: cache.timestamp))
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -26,17 +28,63 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
-		fatalError("Must be implemented")
+		context.perform { [cacheStore, feedImageStore] in
+			do {
+				let cache = try cacheStore.existingCacheOrNewOne()
+				cache.feedImages = feed.map { feedImageStore.feedImage(from: $0) }
```
thiagopenna, line 54: Following from my previous comment, this mapping logic should also be extracted to one of the helpers. It's possible to create a function of type `([LocalFeedImage]) -> [ManagedFeedImage]` in `ManagedFeedImageStore`, and move the mapping logic there:
```suggestion
				cache.feedImages = feedImageStore.images(from: feed)
```

FeedStoreChallenge/Entities/ManagedCache+CoreDataProperties.swift
```diff
@@ -0,0 +1,60 @@
...
+//  Created by Istvan Private on 14.06.21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+//
+
+import Foundation
+import CoreData
+
+@objc(ManagedCache)
+public class ManagedCache: NSManagedObject {
```
thiagopenna, line 14: `ManagedCache` is an implementation detail of `CoreDataFeedStore`, so it should not be marked as `public`. Also, this class is not supposed to be subclassed, so it should be marked as `final`.
```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/Entities/ManagedCache+CoreDataProperties.swift
```diff
@@ -0,0 +1,60 @@
...
+	@nonobjc public class func fetchRequest() -> NSFetchRequest<ManagedCache> {
+		return NSFetchRequest<ManagedCache>(entityName: "ManagedCache")
+	}
+
+	@NSManaged public var timestamp: Date
+	@NSManaged private var feedImagesOrderedSet: NSOrderedSet
+
+	public var feedImages: [ManagedFeedImage] {
```
thiagopenna, line 22: Since this class should not be public, its functions and methods should be changed also.
```suggestion
	@nonobjc class func fetchRequest() -> NSFetchRequest<ManagedCache> {
		return NSFetchRequest<ManagedCache>(entityName: "ManagedCache")
	}

	@NSManaged var timestamp: Date
	@NSManaged private var feedImagesOrderedSet: NSOrderedSet

	var feedImages: [ManagedFeedImage] {
```

FeedStoreChallenge/Entities/ManagedCache+CoreDataProperties.swift
```diff
@@ -0,0 +1,60 @@
...
+		get { feedImagesOrderedSet.array as? [ManagedFeedImage] ?? [] }
```
thiagopenna, line 23: A better approach when dealing with a possible heterogeneous `NSOrderedSet` would be to use a `compactMap`, so all objects that can be downcasted to `ManagedFeedImage` are returned instead of an empty array in case of failure from a single one.
```suggestion
		get { feedImagesOrderedSet.compactMap { $0 as? ManagedFeedImage } }
```

FeedStoreChallenge/Entities/ManagedCache+CoreDataProperties.swift
```diff
@@ -0,0 +1,60 @@
...
+	@NSManaged public func removeFromFeedImagesOrderedSet(_ value: ManagedFeedImage)
+
+	@objc(addFeedImagesOrderedSet:)
+	@NSManaged public func addToFeedImagesOrderedSet(_ values: NSOrderedSet)
+
+	@objc(removeFeedImagesOrderedSet:)
+	@NSManaged public func removeFromFeedImagesOrderedSet(_ values: NSOrderedSet)
+}
+
+extension ManagedCache: Identifiable {}
```
thiagopenna, line 60: These extensions are not used, so they can be removed
```suggestion
```

FeedStoreChallenge/Entities/ManagedFeedImage+CoreDataProperties.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Created by Istvan Private on 14.06.21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+//
+
+import Foundation
+import CoreData
+
+@objc(ManagedFeedImage)
+public class ManagedFeedImage: NSManagedObject {
```
thiagopenna, line 14: `ManagedCache` is an implementation detail of `CoreDataFeedStore`, so it should not be marked as `public`. Also, this class is not supposed to be subclassed, so it should be marked as `final`.
```suggestion
final class ManagedFeedImage: NSManagedObject {
```

FeedStoreChallenge/Entities/ManagedFeedImage+CoreDataProperties.swift
```diff
@@ -0,0 +1,26 @@
...
+	@nonobjc public class func fetchRequest() -> NSFetchRequest<ManagedFeedImage> {
+		return NSFetchRequest<ManagedFeedImage>(entityName: "ManagedFeedImage")
+	}
+
+	@NSManaged public var id: UUID
+	@NSManaged public var descriptionString: String?
+	@NSManaged public var location: String?
+	@NSManaged public var url: URL
+	@NSManaged public var cache: ManagedCache?
```
thiagopenna, line 23: Since this class should not be public, its functions and methods should be changed also.
```suggestion
	@nonobjc class func fetchRequest() -> NSFetchRequest<ManagedFeedImage> {
		return NSFetchRequest<ManagedFeedImage>(entityName: "ManagedFeedImage")
	}

	@NSManaged var id: UUID
	@NSManaged var descriptionString: String?
	@NSManaged var location: String?
	@NSManaged var url: URL
	@NSManaged var cache: ManagedCache?
```

FeedStoreChallenge/Entities/ManagedFeedImage+CoreDataProperties.swift
```diff
@@ -0,0 +1,26 @@
...
+}
+
+extension ManagedFeedImage: Identifiable {}
```
thiagopenna, line 26: This extension is not used, so it can be removed.
```suggestion
```

thiagopenna (commented):
Hi @steven851007,

- All tests are passing ✅ 

Great job!

FeedStoreChallenge/EntityStores/ManagedCacheStore.swift
thiagopenna: Unfortunately, our CI doesn't support Xcode 12.5 because it requires MacOS11, but GitHub Actions support for it is not publicly available yet.

On Xcode 12.4 that line would fail because it wasn't able to disambiguate `self`, and would show as error of `variable used within its own initial value`. One way of using a variable with the same name as another property or method in that context would be to manually disambiguate:
```swift
let fetchRequest = self.fetchRequest()
```

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
thiagopenna: Hi Istvan! Yes, if you depend on the `Cascade` behavior to delete the objects, you can add a test for it. 

Since this is an implementation-specific detail, we don't provide this test for you. But you can create it. For example, by instantiating a core data stack in a new test method and querying the count of image entities (it should be zero after deletion!) ✅

PR #161, opened 2021-06-19
Lívia Vasconcelos - Feed Store Challenge

liuvasconcelos:
Hi, I've finished Feed Store Challenge. Could you review it? Thanks!

### DIFF
--- a/FeedStoreChallenge/Cache.swift
+++ b/FeedStoreChallenge/Cache.swift
@@ -18,4 +18,11 @@ extension Cache {
 	internal var localFeed: [LocalFeedImage] {
 		return feed.compactMap { ($0 as? CacheImage)?.local }
 	}
+
+	internal static func find(in context: NSManagedObjectContext) throws -> Cache? {
+		guard let cacheEntityName = entity().name else { return nil }
+		let request = NSFetchRequest<Cache>(entityName: cacheEntityName)
+		request.returnsObjectsAsFaults = false
+		return try context.fetch(request).first
+	}
 }
--- a/FeedStoreChallenge/Cache.swift
+++ b/FeedStoreChallenge/Cache.swift
@@ -25,4 +25,9 @@ extension Cache {
 		request.returnsObjectsAsFaults = false
 		return try context.fetch(request).first
 	}
+
+	internal static func newUniqueInstance(in context: NSManagedObjectContext) throws -> Cache {
+		try find(in: context).map(context.delete)
+		return Cache(context: context)
+	}
 }
--- a/FeedStoreChallenge/Cache.swift
+++ b/FeedStoreChallenge/Cache.swift
@@ -1,0 +1,15 @@
+//
+//  Cache.swift
+//  FeedStoreChallenge
+//
+//  Created by Livia Vasconcelos on 19/06/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(Cache)
+internal class Cache: NSManagedObject {
+	@NSManaged internal var timestamp: Date
+	@NSManaged internal var feed: NSOrderedSet
+}
--- a/FeedStoreChallenge/Cache.swift
+++ b/FeedStoreChallenge/Cache.swift
@@ -13,3 +13,9 @@ internal class Cache: NSManagedObject {
 	@NSManaged internal var timestamp: Date
 	@NSManaged internal var feed: NSOrderedSet
 }
+
+extension Cache {
+	internal var localFeed: [LocalFeedImage] {
+		return feed.compactMap { ($0 as? CacheImage)?.local }
+	}
+}
--- a/FeedStoreChallenge/CacheImage.swift
+++ b/FeedStoreChallenge/CacheImage.swift
@@ -21,4 +21,15 @@ extension CacheImage {
 	internal var local: LocalFeedImage {
 		return LocalFeedImage(id: id, description: imageDescription, location: location, url: url)
 	}
+
+	internal static func images(from localFeed: [LocalFeedImage], in context: NSManagedObjectContext) -> NSOrderedSet {
+		return NSOrderedSet(array: localFeed.map { local in
+			let managed = CacheImage(context: context)
+			managed.id = local.id
+			managed.imageDescription = local.description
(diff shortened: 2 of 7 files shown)


FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="17709" systemVersion="20C69" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="Cache" representedClassName="Cache" syncable="YES">
+        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feed" optional="YES" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="CacheImage" inverseName="cache" inverseEntity="CacheImage"/>
```
thiagopenna, line 5: The `FeedStore.xcdatamodeld` declares `Cache`'s `feed` relationship as `optional`, but the `@NSManaged var feed` property is declared as a non-optional `NSOrderedSet`. This could lead to runtime issues since having a `Cache` without a related feed is a representable state.

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
...
+    </entity>
+    <entity name="CacheImage" representedClassName="CacheImage" syncable="YES">
+        <attribute name="id" attributeType="UUID" usesScalarValueType="NO"/>
+        <attribute name="imageDescription" optional="YES" attributeType="String"/>
+        <attribute name="location" optional="YES" attributeType="String"/>
+        <attribute name="url" attributeType="URI"/>
+        <relationship name="cache" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="Cache" inverseName="feed" inverseEntity="Cache"/>
```
thiagopenna, line 12: The `FeedStore.xcdatamodeld` declares `CacheImage`'s `cache` relationship as `optional`, but the `@NSManaged var cache` property is declared as a non-optional `Cache`. This could lead to runtime issues since having a `CacheImage` without a related cache is a representable state.

FeedStoreChallenge/CacheImage.swift
```diff
@@ -0,0 +1,35 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Livia Vasconcelos on 19/06/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(CacheImage)
+internal class CacheImage: NSManagedObject {
```
thiagopenna, line 12: This class name can be confusing, since it can be understood as "One image that represents one specific `Cache` object" (similar to how you would read "UserImage", for example). To better clarify it's meaning I suggest changing it to `CachedImage`

thiagopenna: The convention in this codebase is to omit the default access level keyword of `internal`, please remove it from this class and its properties and methods.
Also, `CacheImage` is not supposed to be subclassed, so it should be marked as `final`.
```suggestion
final class CacheImage: NSManagedObject {
```

thiagopenna (commented):
Hi @liuvasconcelos ,

- All tests are passing ✅ 

I left some review comments, feel free to push changes.

Great job!

FeedStoreChallenge/Cache.swift
```diff
@@ -0,0 +1,33 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Livia Vasconcelos on 19/06/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(Cache)
+internal class Cache: NSManagedObject {
```
thiagopenna, line 12: The convention in this codebase is to omit the default access level keyword of `internal`, please remove it from this class and its properties and methods.
Also, `Cache` is not supposed to be subclassed, so it should be marked as `final`.
```suggestion
final class Cache: NSManagedObject {
```

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
liuvasconcelos: That's true! Fixed, thanks!

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
liuvasconcelos: That's true! Fixed, thanks!

FeedStoreChallenge/CacheImage.swift
liuvasconcelos: Renamed and fixed!

FeedStoreChallenge/Cache.swift
liuvasconcelos: Fixed!

liuvasconcelos:
@thiagopenna I fixed all your comments, could you take a look to see if everythink is ok now? Thanks!

FeedStoreChallenge/CacheImage.swift
thiagopenna: Regarding the `internal` keyword, I believe you missed the part where I said "please remove it from this class **and its properties and methods**."
I preferred to not make a gigantic suggestion block with all of them, but I should have been clearer.

FeedStoreChallenge/Cache.swift
thiagopenna: Same as the `CachedImage`, please remove the `internal` keyword from properties and methods of `Cache`.

thiagopenna:
Hi @liuvasconcelos , I unresolved two comments regarding the `internal` keyword, feel free to push changes.

PR #170, opened 2021-07-25
Luis Zapata - Feed Store Challenge

zgluis:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,53 @@ public final class CoreDataFeedStore: FeedStore {
...
+						try self.context.save()
+						completion(nil)
+					} catch {
+						completion(error)
+					}
+				}
+			} else {
+				completion(error)
+			}
+		}
```
danillahtin, line 63: Calling `deleteCachedFeed` first will make insertion less performant as it would be separated into two chunks of code to run on the context's internal queue. Thus in this case it is better to delete previously stored cache manually and have only one `context.perform` block.

Also using `self` inside this block captures self, which is not needed. To avoid it `context` that was created on line 48 can be used. 

```suggestion
		let context = context
		context.perform {
			do {
				try context.fetch(ManagedCache.fetchRequest()).first.map(context.delete)
				
				let _ = ManagedCache(context: context, feed: feed, timestamp: timestamp)

				try context.save()
				completion(nil)
			} catch {
				completion(error)
			}
		}
```

Also it helps to avoid callback hell and keeps the code concise

danillahtin (commented):
Hi @zgluis, you almost passed the challenge!

Thank you 🙂

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,53 @@ public final class CoreDataFeedStore: FeedStore {
...
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		let context = context
+		context.performAndWait {
+			let fetchRequest: NSFetchRequest = ManagedCache.fetchRequest()
+			do {
+				let result = try self.context.fetch(fetchRequest)
+				result.first.map(context.delete)
+				try self.context.save()
+				completion(nil)
```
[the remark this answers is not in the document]
zgluis: Awesome suggestions thanks! @danillahtin

FeedStoreChallenge/CoreDataFeedStore.swift
zgluis: Thanks! I'm applying the changes however I have some doubts:

1. At first I implemented it the way you mentioned but then I thought that I was duplicating the `delete` functionality inside `insert` so I decided to start calling `delete`, that created the nested closure but I felt that I was not "repeating" code. Does that make any sense for you?.

2. I also tried to do batch delete, but the tests were unable to inject errors properly, I read some docs and found that it is not easy to handle the result of a batch delete. Is there a way to use it here? @danillahtin

FeedStoreChallenge.xcodeproj/project.pbxproj
```diff
@@ -21,6 +21,10 @@
 		08F35C742616135B00AFE277 /* XCTestCase+MemoryLeakTracking.swift in Sources */ = {isa = PBXBuildFile; fileRef = 08F35C732616135B00AFE277 /* XCTestCase+MemoryLeakTracking.swift */; };
 		08F35C7A2616166600AFE277 /* XCTestCase+FailableRetrieveFeedStoreSpecs.swift in Sources */ = {isa = PBXBuildFile; fileRef = 08F35C772616166600AFE277 /* XCTestCase+FailableRetrieveFeedStoreSpecs.swift */; };
 		08F35C7B2616166600AFE277 /* XCTestCase+FailableInsertFeedStoreSpecs.swift in Sources */ = {isa = PBXBuildFile; fileRef = 08F35C782616166600AFE277 /* XCTestCase+FailableInsertFeedStoreSpecs.swift */; };
+		2818933426A37AEF006E0DA4 /* (null) in Sources */ = {isa = PBXBuildFile; };
```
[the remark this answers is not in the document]
zgluis: Thanks a lot for the comments @danillahtin , I addressed all of them, please check

FeedStoreChallenge/CoreDataFeedStore.swift
danillahtin: Hey @zgluis!

1. Yes, I got that from the code. The thing is that duplication is gone but other problems arose (performance, callbacks, bad readability). For this function you only need to remove stored `ManagedCache` if it was persisted. `deleteCachedFeed` not only deletes `ManagedCache`, but also it saves the context. Actually it is a method of another level of abstraction. Potentially in the future `deleteCachedFeed` can appear to have a lot more details which would be redundant or even harmful to use inside insert. There's a way to get rid of duplication without reusing higher level functions: create a helper that will do only the deletion, that's it. Thus you will be able to reuse it everywhere without all the cons of reusing `deleteCachedFeed`.
2. I don't see batch deletion here. We know that `CoreDataFeedStore` operates only on one object. The tests guarantee that we do it properly so that there're no 2 instances of `ManagedCache` at a time in Core Data. Thus it is more efficient to do the minimal amount of work and not to try remove 1 object with batch deletion.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,48 @@ public final class CoreDataFeedStore: FeedStore {
 	}
 
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		fatalError("Must be implemented")
+		let context = context
+		context.perform {
```
danillahtin, line 33: Just FYI, there's another way of capturing `context` inside the block so that `self` is not captured. One can use block's capture list. The following suggestion does 100% the same but 1 line shorter

```suggestion
		context.perform { [context] in
```

danillahtin (commented):
Hi @zgluis,

- All tests are passing ✅ 

Great job!

PR #172, opened 2021-07-30
Ankit Kumar - Feed Store Challenge

ankitkumar-ios:


FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,34 @@
...
+	@NSManaged internal var feed_description: String?
+	@NSManaged internal var location: String?
+	@NSManaged internal var url: URL
+	@NSManaged internal var cache: ManagedCache
+
+	var local: LocalFeedImage {
+		return LocalFeedImage(id: id, description: feed_description, location: location, url: url)
+	}
+
+	internal static func image(from localFeed: [LocalFeedImage], in context: NSManagedObjectContext) -> NSOrderedSet {
```
danillahtin, line 23: `internal` is a default access level, no need to declare it explicitly. Please remove `internal` modifier.
Also this method returns a collection so the name can be in plural to better describe intent

```suggestion
	static func images(from localFeed: [LocalFeedImage], in context: NSManagedObjectContext) -> NSOrderedSet {
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,30 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Ankit on 30/07/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedCache)
+internal class ManagedCache: NSManagedObject {
```
danillahtin, line 12: `internal` is a default access level, no need to declare it explicitly.
Also this class is not intended to be subclassed thus it should be `final`.

Please remove `internal` modifier and make the class `final`.
```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,30 @@
...
+	internal static func newUniqueInstance(in context: NSManagedObjectContext) throws -> ManagedCache {
+		try find(in: context).map(context.delete)
+		return ManagedCache(context: context)
+	}
+
+	internal static func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		let request = NSFetchRequest<ManagedCache>(entityName: ManagedCache.entity().name!)
+		request.returnsObjectsAsFaults = false
+		return try context.fetch(request).first
+	}
```
danillahtin, line 29: `internal` is a default access level, no need to declare it explicitly. Please remove `internal` modifiers.

```suggestion
	@NSManaged var timestamp: Date
	@NSManaged var feed: NSOrderedSet

	var localFeed: [LocalFeedImage] {
		return feed.compactMap { ($0 as? ManagedFeedImage)?.local }
	}

	static func newUniqueInstance(in context: NSManagedObjectContext) throws -> ManagedCache {
		try find(in: context).map(context.delete)
		return ManagedCache(context: context)
	}

	static func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
		let request = NSFetchRequest<ManagedCache>(entityName: ManagedCache.entity().name!)
		request.returnsObjectsAsFaults = false
		return try context.fetch(request).first
	}
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,50 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				try ManagedCache.find(in: context).map(context.delete)
+				try context.save()
```
danillahtin, line 65: In case there is no previously stored cache `context.delete` will not change the context, thus there will be no changes in it and will be no need to save it. Otherwise, Core Data may perform unnecessary work.

One way to ensure the context is only saved when there are changes is to use `.map`:
```suggestion
				try ManagedCache.find(in: context).map(context.delete).map(context.save)
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,50 @@ public final class CoreDataFeedStore: FeedStore {
...
+			} catch {
+				context.rollback()
+				completion(error)
+			}
+		}
+	}
+
+	private func perform(_ action: @escaping (NSManagedObjectContext) -> Void) {
+		let context = self.context
+		context.perform { action(context) }
```
danillahtin, line 76: During the course this way of having a local reference to the context to avoid capturing self inside of the `context.perform` block is taught.
The same can be achieved using capture list.

```suggestion
		context.perform { [context] in action(context) }
```

danillahtin (commented):
Hi @ankitkumar-ios, you almost passed the challenge!

Thank you 🙂

ankitkumar-ios:
Hi Danil, 
Thanks for the feedback. I have added all there required changed

FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,34 @@
...
+
+import CoreData
+
+@objc(ManagedFeedImage)
+final class ManagedFeedImage: NSManagedObject {
+	@NSManaged var id: UUID
+	@NSManaged var feed_description: String?
+	@NSManaged var location: String?
+	@NSManaged var url: URL
+	@NSManaged var cache: ManagedCache
```
danillahtin, line 17: `cache` is marked as optional in `FeedStore.xcdatamodel`. Please make it optional here so that the class and the model are consistent

```suggestion
	@NSManaged var cache: ManagedCache?
```

danillahtin (commented):
Hi @ankitkumar-ios, almost there!

Please, check my review comment and let me know once you're done so I can review it again. 

Thank you 🙂

ankitkumar-ios:
Hi Danil,
Made the required changes and pushed the code. Please review
Thanks

danillahtin (commented):
Hi @ankitkumar-ios,

- All tests are passing ✅ 

Great job!

PR #175, opened 2021-08-12
Bryan Hoke - Feed Store Challenge

BryanHoke:


FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,35 @@
...
+
+import CoreData
+
+@objc(ManagedFeedImage)
+class ManagedFeedImage: NSManagedObject {
+	@NSManaged var id: UUID
+	@NSManaged var imageDescription: String?
+	@NSManaged var location: String?
+	@NSManaged var url: URL
+	@NSManaged var cache: ManagedCache
```
thiagopenna, line 17: The `FeedStore.xcdatamodeld` declares `ManagedFeedImage`'s `cache` relationship as `optional`, but the `@NSManaged var cache` property is declared as a non-optional `ManagedCache`. This could lead to runtime issues since having a `ManagedFeedImage` without a related cache is a representable state.

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,38 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Bryan Hoke on 8/11/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedCache)
+class ManagedCache: NSManagedObject {
```
thiagopenna, line 12: `ManagedCache` is not supposed to be subclassed, so it should be marked as `final`:
```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,35 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Bryan Hoke on 8/11/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedFeedImage)
+class ManagedFeedImage: NSManagedObject {
```
thiagopenna, line 12: `ManagedFeedImage` is not supposed to be subclassed, so it should be marked as `final`:
```suggestion
final class ManagedFeedImage: NSManagedObject {
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,38 @@
...
+	@NSManaged var feed: NSOrderedSet
+}
+
+extension ManagedCache {
+	static func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		guard let entityName = ManagedCache.entity().name else {
+			return nil
+		}
+		let request = NSFetchRequest<ManagedCache>(entityName: entityName)
+		return try context.fetch(request).first
```
thiagopenna, line 23: To slightly increase CoreData performance `request.returnsObjectsAsFaults` can be set to `false`.

```suggestion
		let request = NSFetchRequest<ManagedCache>(entityName: entityName)
		request.returnsObjectsAsFaults = false
		return try context.fetch(request).first
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -26,17 +26,52 @@ public final class CoreDataFeedStore: FeedStore {
 			url: storeURL
 		)
 		context = container.newBackgroundContext()
+		context.undoManager = UndoManager()
```
thiagopenna, line 29: There's no need to set an `UndoManager` since we are not using `undo()` messages. Also, according to the [documentation](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506663-undomanager):
> _You can set the undo manager to nil to disable undo support. This provides a performance benefit if you do not want to support undo for a particular context [...]_
_[...] on iOS, the undo manager is nil by default._

Setting an `undoManager` in this case will simply hurt perfomance while bringing no benefits.
```suggestion
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -26,17 +26,52 @@ public final class CoreDataFeedStore: FeedStore {
...
 	}
+}
 
+extension CoreDataFeedStore: FeedStore {
 	public func retrieve(completion: @escaping RetrievalCompletion) {
-		fatalError("Must be implemented")
+		let context = self.context
+		context.perform {
```
thiagopenna, line 36: Suggestion: another way of capturing `context` to use inside this closure is by using a Capture List:
```suggestion
		context.perform { [context] in
```
If you decide to follow this suggestion, apply it to the other `context.perform` calls in this file to keep the code consistent.

thiagopenna (commented):
Hi @BryanHoke , you almost passed the challenge!

Thank you 🙂

FeedStoreChallenge/ManagedFeedImage.swift
BryanHoke: Thanks, good catch!

BryanHoke:
Hello @thiagopenna,

Thanks for your review! I have made changed based on your review comments and am ready for review again.

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -27,16 +27,48 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		context.perform { [context] in
+			do {
+				try ManagedCache.deleteCache(in: context)
+				try context.save()
```
thiagopenna, line 66: In case there is no previously stored cache `context.delete` will not change the context, thus there will be no changes in it and will be no need to save it. Otherwise, Core Data may perform unnecessary work.

One way to ensure the context is only saved when there are changes is to use `context.hasChanges`:
```suggestion
				try ManagedCache.deleteCache(in: context)
				if context.hasChanges {
					try context.save()
				}
```

thiagopenna (commented):
Hi @BryanHoke,

✅  All tests are passing 

I left one final review comment. Feel free to push changes.

PR #176, opened 2021-08-12
Stanislav Li - Feed Store Challenge 

gitlees:


FeedStoreChallenge/CoreData/ManagedCache.swift
```diff
@@ -0,0 +1,28 @@
...
+
+	var localFeed: [LocalFeedImage] {
+		return feed.compactMap { ($0 as? ManagedFeedImage)?.asLocal }
+	}
+}
+
+extension ManagedCache {
+	static func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		let request = NSFetchRequest<ManagedCache>(entityName: ManagedCache.entity().name!)
+		return try context.fetch(request).first
```
thiagopenna, line 26: To slightly increase CoreData performance `request.returnsObjectsAsFaults` can be set to `false`.

```suggestion
		let request = NSFetchRequest<ManagedCache>(entityName: entity().name!)
		request.returnsObjectsAsFaults = false
		return try context.fetch(request).first
```

FeedStoreChallenge/CoreData/ManagedFeedImage.swift
```diff
@@ -0,0 +1,37 @@
...
+	@NSManaged var cache: ManagedCache?
+
+	var asLocal: LocalFeedImage {
+		return LocalFeedImage(id: id, description: imageDescription, location: imageLocation, url: url)
+	}
+}
+
+extension Array where Element == LocalFeedImage {
+	func asOrderedSet(in context: NSManagedObjectContext) -> NSOrderedSet {
+		return NSOrderedSet(array: map { local in
```
thiagopenna, line 28: This helper's name is misleading, since it does more than simply converting the `Array` instance to an `OrderedSet`: it also converts the `LocalFeedImages` to `ManagedFeedImages`, and that's not apparent to clients of this method.

Consider renaming it to make it clearer when using, like `toManagedImages` (the `OrderedSet` part can be omitted since it's obvious from the return type):
```suggestion
extension Array where Element == LocalFeedImage {
	func toManagedImages(in context: NSManagedObjectContext) -> NSOrderedSet {
		return NSOrderedSet(array: map { local in
```
That way it will read as:
```swift
managedCache.feed = feed.toManagedImages(in: context)
```
Also, consider moving this extension to where it's being used and setting it as `private`, since it's an implementation detail of how the `ManagedCache` is structured and should not be accessible from the whole module.

--- 

Another option would be to create a static helper method inside `ManagedFeedImage`, that receives an array of `LocalFeedImage`, and returns a `NSOrderedSet` using this extension (so we can mark it as private and leave it in this file). That way, we could call it as:
```swift
managedCache.feed = ManagedFeedImage.images(from: feed, in: context)
```

FeedStoreChallenge/CoreData/ManagedCache.swift
```diff
@@ -0,0 +1,28 @@
...
+//  Created by Stas Lee on 11/8/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+//
+
+import Foundation
+import CoreData
+
+@objc(ManagedCache)
+class ManagedCache: NSManagedObject {
```
thiagopenna, line 14: `ManagedCache` is not supposed to be subclassed, so it should be marked as `final`:
```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/CoreData/ManagedFeedImage.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Created by Stas Lee on 11/8/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+//
+
+import Foundation
+import CoreData
+
+@objc(ManagedFeedImage)
+class ManagedFeedImage: NSManagedObject {
```
thiagopenna, line 14: `ManagedFeedImage` is not supposed to be subclassed, so it should be marked as `final`:
```suggestion
final class ManagedFeedImage: NSManagedObject {
```

thiagopenna (commented):
Hi @gitlees, you almost passed the challenge!

Thank you 🙂

FeedStoreChallenge/CoreData/ManagedCache.swift
gitlees: Got a question here. Not supported to subclasses in terms of CoreData models as dto ? Thus we are restricting the subclassing functionality here @thiagopenna

gitlees:
Thank you for the review ! Great considerations to think of.

Done @thiagopenna

FeedStoreChallenge/CoreData/ManagedCache.swift
thiagopenna: Hey @gitlees, I mean that based on our current design, there's no immediate need to subclass `ManagedCache`. Knowing that, we can increase performance by reducing dynamic dispatch, using the `final` keyword. From Apple's Swift [Blog](https://developer.apple.com/swift/blog/?id=27):

> [...] dynamic dispatch, increases language expressivity at the cost of a constant amount of runtime overhead for each indirect usage. In performance sensitive code such overhead is often undesirable. [...]
>
> The final keyword is a restriction on a class, method, or property that indicates that the declaration cannot be overridden. This allows the compiler to safely elide dynamic dispatch indirection. [...]

If we need to subclass in the future (for a good reason), we can remove this optimization.

thiagopenna (commented):
Hi @gitlees,

✅  All tests are passing 

I left an answer for the question you asked from the previous comments.

PR #177, opened 2021-08-15
Beniamin Idziak - Feed Store Challenge

beniaminidziak:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,51 @@ public final class CoreDataFeedStore: FeedStore {
...
+			} catch {
+				context.rollback()
+				completion(error)
+			}
+		}
+	}
+
+	private func perform(_ action: @escaping (NSManagedObjectContext) -> Void) {
+		let context = context
+		context.perform {
```
caiozullo, line 75: Another way of capturing the `context` without capturing `self` inside of the `perform` block is to use a capture list:

```suggestion
		context.perform { [context] in
```

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="18154" systemVersion="20F71" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="ManagedCache" representedClassName="ManagedCache" syncable="YES">
+        <attribute name="timestamp" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
```
caiozullo, line 4: `timestamp` shouldn't be optional:

```suggestion
        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
```

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
...
+        <relationship name="feed" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
+    </entity>
+    <entity name="ManagedFeedImage" representedClassName="ManagedFeedImage" syncable="YES">
+        <attribute name="id" attributeType="UUID" usesScalarValueType="NO"/>
+        <attribute name="imageDescription" optional="YES" attributeType="String"/>
+        <attribute name="location" optional="YES" attributeType="String"/>
+        <attribute name="url" optional="YES" attributeType="URI"/>
```
caiozullo, line 11: `url` shouldn't be optional:

```suggestion
        <attribute name="url" attributeType="URI"/>
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	static func delete(in context: NSManagedObjectContext) throws {
+		try cache(in: context).map(context.delete).map(context.save)
+	}
+
+	static func new(in context: NSManagedObjectContext) throws -> ManagedCache {
+		try delete(in: context)
+		return ManagedCache(context: context)
+	}
```
caiozullo, line 37: The `delete` method calls `context.save`, which performs a disk write. But then, we create a new `ManagedCache` and perform another save operation. Calling `save` is an expensive operation, so we should avoid calling it multiple times when not needed.

Ideally, those operations should happen in a single call to `context.save`. Otherwise, Core Data may perform extra work or if the second save fails, the database will be in an inconsistent state (it deleted the previous cache but didn't add the new one).

I recommend you remove the `.map(context.save)` from the `delete` static method to avoid these problems. You can perform the save operation in the `CoreDataFeedStore` class where appropriate.

caiozullo (commented):
Hi @beniaminidziak, you almost passed the challenge!

Thank you 🙂

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+//  Created by Beniamin Idziak on 15/08/2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+@objc(ManagedCache)
+class ManagedCache: NSManagedObject {
```
caiozullo, line 13: This class is not intended to be subclassed, so it can be final:

```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+//  Created by Beniamin Idziak on 15/08/2021.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import Foundation
+import CoreData
+
+@objc(ManagedFeedImage)
+class ManagedFeedImage: NSManagedObject {
```
caiozullo, line 13: This class is not intended to be subclassed, so it can be final:

```suggestion
final class ManagedFeedImage: NSManagedObject {
```

FeedStoreChallenge/CoreDataFeedStore.swift
beniaminidziak: I like this kind of 'syntactic sugar' thanks :)

beniaminidziak:
@caiozullo Thanks for review 😄 recommended changes are already in PR

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,52 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				try ManagedCache.delete(in: context)
+				try context.save()
```
caiozullo, line 65: In case the cache is empty, `ManagedCache.delete` will not delete anything or change the managed context. Thus, there will be no need to call `save`. 

Core Data may perform unnecessary work when calling save, even if there are no changes.

You can check the `NSManagedObjectContext.hasChanges`property before calling save to avoid unnecessary work:

```suggestion
				try ManagedCache.delete(in: context)
				if context.hasChanges { 
					try context.save()
				}
```

caiozullo (commented):
Hi @beniaminidziak,

✅  All tests are passing 

**I left a final review comment. Feel free to push changes.**

PR #179, opened 2021-08-17
Daniel Gallego Peralta - Feed Store Challenge

danielios-hub:


FeedStoreChallenge/CoreData/Models/ManagedFeedImage.swift
```diff
@@ -0,0 +1,36 @@
...
+
+import CoreData
+
+@objc(ManagedFeedImage)
+internal class ManagedFeedImage: NSManagedObject {
+	@NSManaged var id: UUID
+	@NSManaged var imageDescription: String?
+	@NSManaged var location: String?
+	@NSManaged var url: URL
+	@NSManaged var cache: NSManagedObject
```
danillahtin, line 17: `cache` property has type of `ManagedCache` in `FeedStore.xcdatamodel`. Please change its type here so that model and class declaration are consistent.

```suggestion
	@NSManaged var cache: ManagedCache
```

FeedStoreChallenge/CoreData/Models/ManagedFeedImage.swift
```diff
@@ -0,0 +1,36 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Daniel Gallego Peralta on 17/8/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedFeedImage)
+internal class ManagedFeedImage: NSManagedObject {
```
danillahtin, line 12: `internal` is the default access level, no need to declare it explicitly.
Please remove `internal` modifier and make the class `final`.

```suggestion
final class ManagedFeedImage: NSManagedObject {
```

FeedStoreChallenge/CoreData/Models/ManagedCache.swift
```diff
@@ -0,0 +1,32 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Daniel Gallego Peralta on 17/8/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedCache)
+internal class ManagedCache: NSManagedObject {
```
danillahtin, line 12: `internal` is a default access level, no need to declare it explicitly.
Please remove `internal` modifier and make the class `final`.

```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/CoreData/Models/ManagedCache.swift
```diff
@@ -0,0 +1,32 @@
...
+		}
+	}
+
+	static func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		let request = NSFetchRequest<ManagedCache>(entityName: ManagedCache.entity().name!)
+		request.returnsObjectsAsFaults = false
+		return try context.fetch(request).first
+	}
+
+	internal static func newUniqueInstance(in context: NSManagedObjectContext) throws -> ManagedCache {
```
danillahtin, line 28: `internal` is the default access level, so you don't need to write it explicitly.

It was written explicitly in the first lectures of the program for teaching purposes, but you don't need to do it in your code.

Please remove `internal` modifier

```suggestion
	static func newUniqueInstance(in context: NSManagedObjectContext) throws -> ManagedCache {
```

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,54 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				try ManagedCache.find(in: context).map(context.delete)
+				try context.save()
```
danillahtin, line 67: In case there is no previously stored cache `context.delete` will not change the context, thus there will be no changes in it and will be no need to save it. Otherwise, Core Data may perform unnecessary work.
To ensure the context is only saved when there are change optional chaining can be used:

```suggestion
				try ManagedCache.find(in: context)
				.map(context.delete)
				.map(context.save)
```

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,54 @@ public final class CoreDataFeedStore: FeedStore {
...
 	}
 
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				let feed = ManagedFeedImage.images(from: feed, in: context)
+				let cache = try ManagedCache.newUniqueInstance(in: context)
+				cache.timestamp = timestamp
+				cache.feed = feed
```
danillahtin, line 52: Suggestion:
There's no need to convert `feed` to images before `try ManagedCache.newUniqueInstance(in: context)` because try can fail and unnecessary work can be performed.
Also setting `cache.feed` can be a one-liner.

```suggestion
				let cache = try ManagedCache.newUniqueInstance(in: context)
				cache.timestamp = timestamp
				cache.feed = ManagedFeedImage.images(from: feed, in: context)
```

FeedStoreChallenge/CoreData/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,54 @@ public final class CoreDataFeedStore: FeedStore {
...
+				completion(error)
+			}
+		}
+	}
+
+	func perform(_ action: @escaping (NSManagedObjectContext) -> Void) {
+		let context = self.context
+		context.perform {
+			action(context)
+		}
```
danillahtin, line 80: Suggestion:
During the course this way of having a local reference to the `context` to avoid capturing `self` inside of the `context.perform` block is taught.
The same can be achieved using a capture list.

```suggestion
		context.perform { [context] in 
			action(context) 
		}
```

danillahtin (commented):
Hi @danielios-hub, you almost passed the challenge!

Thank you 🙂

danillahtin:
Hi @danielios-hub. 
Do you need any help performing the changes requested in the review? Just let me know - We're here to help! 🙂

PR #180, opened 2021-08-18
Haik Ampardjian - Feed Store Challenge

ludoded:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,54 @@ public final class CoreDataFeedStore: FeedStore {
...
+				completion(.failure(error))
+			}
+		}
 	}
 
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				try? ManagedCache.find(in: context).map(context.delete)
```
thiagopenna, line 50: If `ManagedCache.find` throws any errors, they should not be ignored with optional `try?`. Otherwise, we may leave the store in an invalid state with multiple caches, if the `delete` failed, but `insert` works.
```suggestion
				try ManagedCache.find(in: context).map(context.delete)
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,54 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				try ManagedCache.find(in: context).map(context.delete)
+				try context.save()
```
thiagopenna, line 69: In case there is no previously stored cache `context.delete` will not change the context, thus there will be no changes in it and will be no need to save it. Otherwise, Core Data may perform unnecessary work.

One way to ensure the context is only saved when there are changes is to use `.map`:
```suggestion
				try ManagedCache.find(in: context).map(context.delete).map(context.save)
```

FeedStoreChallenge/Model/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
...
+    <entity name="ManagedCache" representedClassName=".ManagedCache" syncable="YES">
+        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feed" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
+    </entity>
+    <entity name="ManagedFeedImage" representedClassName=".ManagedFeedImage" syncable="YES">
+        <attribute name="id" attributeType="UUID" usesScalarValueType="NO"/>
+        <attribute name="imageDescription" optional="YES" attributeType="String"/>
+        <attribute name="location" optional="YES" attributeType="String"/>
+        <attribute name="url" attributeType="URI"/>
+        <relationship name="cache" maxCount="1" deletionRule="Cascade" destinationEntity="ManagedCache" inverseName="feed" inverseEntity="ManagedCache"/>
```
thiagopenna, line 12: `ManagedFeedImage`'s `deletionRule` should be `Nullify`, otherwise if one image is deleted, the whole `ManagedCache` would also be deleted.
```suggestion
        <relationship name="cache" maxCount="1" deletionRule="Nullify" destinationEntity="ManagedCache" inverseName="feed" inverseEntity="ManagedCache"/>
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,54 @@ public final class CoreDataFeedStore: FeedStore {
...
+			} catch {
+				context.rollback()
+				completion(error)
+			}
+		}
+	}
+
+	private func perform(action: @escaping (NSManagedObjectContext) -> Void) {
+		let context = self.context
+		context.perform { action(context) }
```
thiagopenna, line 80: Suggestion: 
During the course this way of having a local reference to the `context` to avoid capturing `self` inside of the `context.perform` block is taught.
The same can be achieved using a capture list.

```suggestion
		context.perform { [context] in action(context) }
```

FeedStoreChallenge/Model/ManagedCache.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+import CoreData
+
+class ManagedCache: NSManagedObject {
```
thiagopenna, line 4: `ManagedCache` is not supposed to be subclassed, so it should be marked as `final`:
```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/Model/ManagedFeedImage.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+import CoreData
+
+class ManagedFeedImage: NSManagedObject {
```
thiagopenna, line 4: `ManagedFeedImage` is not supposed to be subclassed, so it should be marked as `final`:
```suggestion
final class ManagedFeedImage: NSManagedObject {
```

thiagopenna (commented):
Hi @ludoded, you almost passed the challenge!

Thank you 🙂

ludoded:
Hi @thiagopenna 
Thanks for your review, those comments are all on point!
I've pushed the fixes through the commits. 

Let me know if you'll find anything else.
Cheers

thiagopenna (commented):
Hi @ludoded,

✅  All tests are passing 

PR #183, opened 2021-08-20
Prabhat Tiwari - Feed Store Challenge

Prabhat07:


FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="17192" systemVersion="19G73" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+    <entity name="ManagedCache" representedClassName="ManagedCache" syncable="YES">
+        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feed" optional="YES" toMany="YES" deletionRule="Nullify" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
```
thiagopenna, line 5: The `FeedStore.xcdatamodeld` declares `ManagedCache`'s `feed` relationship as `optional`, but the `@NSManaged var feed` property is declared as a non-optional `NSOrderedSet`. This could lead to runtime issues since having a `ManagedCache` without a related feed is a representable state.

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
...
+    </entity>
+    <entity name="ManagedFeedImage" representedClassName="ManagedFeedImage" syncable="YES">
+        <attribute name="id" attributeType="UUID" usesScalarValueType="NO"/>
+        <attribute name="imageDescription" optional="YES" attributeType="String"/>
+        <attribute name="location" optional="YES" attributeType="String"/>
+        <attribute name="url" attributeType="URI"/>
+        <relationship name="cache" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="ManagedCache" inverseName="feed" inverseEntity="ManagedCache"/>
```

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
thiagopenna: The `feed` relationship `deletionRule` should be `Cascade` to prevent orphaned objects in the DB (when we delete the `ManagedCache` instance, we want all images in the `feed` property to be deleted as well!).
```suggestion
        <relationship name="feed" optional="YES" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,30 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Prabhat Tiwari on 20/08/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedCache)
+internal class ManagedCache: NSManagedObject {
```
thiagopenna, line 12: The convention in this codebase is to omit the default access level keyword `internal`. Also, `ManagedCache` is not supposed to be subclassed, so it should be marked as `final`:
```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,33 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Prabhat Tiwari on 20/08/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedFeedImage)
+internal class ManagedFeedImage: NSManagedObject {
```
thiagopenna, line 12: Same as my previous comment
```suggestion
final class ManagedFeedImage: NSManagedObject {
```

thiagopenna (commented):
Hi @Prabhat07, you almost passed the challenge!

Thank you 🙂

Prabhat07:
> Hi @Prabhat07, you almost passed the challenge!
> 
> This is Penna, I am an instructor at Essential Developer.
> 
> Please, check my review comments and let me know once you're done so I can review it again.
> 
> Thank you 🙂

Hello @thiagopenna, sorry for delay, made the suggested changes please review again, Thanks!! 😊

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
Prabhat07: Can you please brief what does representable state means?
Thanks!

thiagopenna: Absolutely. By `representable state` I'm referring to any combination of values that our application might be in. I recommend the lesson on [Eliminating Invalid Paths](https://academy.essentialdeveloper.com/courses/447455/lectures/8644863) and how Caio and Mike _make invalid paths unrepresentable_ . when dealing with optionals regarding `HTTPURLResponse`, `Data` and `Error`.

---

The problematic state in this case is that our `xcdatamodeld` allows a `ManagedFeedImage` to be saved without a `cache` relationship (even though the current implementation does not allow the creation of a ManagedFeedImage directly, that might change in the future).

But if we try to load a `ManagedFeedImage` without a cache we will have problems, because the `ManagedFeedImage` model declares our cache as non-optional: `@NSManaged var cache: ManagedCache`. And that's one reason why both the DataModel and the actual class implementation should represent the same entity.

Prabhat07: Make sense, do we also need to change the deletion rule for this?

thiagopenna: Not for the `ManagedFeedImage`: `Nullify` is ok. We don't want to delete the whole `ManagedCache` whenever a single image is deleted.

FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStore.xcdatamodel/contents
thiagopenna: > The `feed` relationship `deletionRule` should be `Cascade` to prevent orphaned objects in the DB (when we delete the `ManagedCache` instance, we want all images in the `feed` property to be deleted as well!).

We should change the deletion rule for `ManagedCache`'s `feed` relationship.

thiagopenna:
Well done!

PR #184, opened 2021-08-21
Valentin Šarić - Feed Store Challenge

VSharkO:


FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,87 @@ public final class CoreDataFeedStore: FeedStore {
...
+				completion(nil)
+			} catch {
+				context.reset()
+				completion(error)
+			}
+		}
 	}
 }
+
+public class ManagedCache: NSManagedObject {
```
danillahtin, line 92: `ManagedCache` is an internal implementation detail of `CoreDataFeedStore`. It doesn't need to be public and since it is declared inside `CoreDataFeedStore.swift` it doesn't need to be `internal` as well and can be `private`. 
Please make the class `private final`.

```suggestion
private final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,87 @@ public final class CoreDataFeedStore: FeedStore {
...
+
+	static func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		guard let name = entity().name else { return nil }
+		let request = NSFetchRequest<ManagedCache>(entityName: name)
+		request.returnsObjectsAsFaults = false
+		return try context.fetch(request).first
+	}
+}
+
+public class ManagedFeedImage: NSManagedObject {
```
danillahtin, line 109: `ManagedFeedImage` should be `private final`, the same way as `ManagedCache` should.

Please make the class `private final`.

```suggestion
private final class ManagedFeedImage: NSManagedObject {
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,87 @@ public final class CoreDataFeedStore: FeedStore {
...
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		let context = self.context
+		context.perform {
+			do {
+				try ManagedCache.find(in: context).map(context.delete).map(context.save)
+				completion(nil)
+			} catch {
+				context.reset()
```
danillahtin, line 85: `context.reset` invalidates all the managed objects of this context which in turn can be an issue if some of these objects are in use. To only reset the changes `context.rollback` needs to be used. 

Please replace `reset` call with `rollback`

```suggestion
				context.rollback()
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,87 @@ public final class CoreDataFeedStore: FeedStore {
...
+					managed.id = local.id
+					managed.imageDescription = local.description
+					managed.location = local.location
+					managed.url = local.url
+					return managed
+				})
+				try context.save()
+				completion(nil)
+			} catch {
+				context.reset()
```
danillahtin, line 72: `context.reset` invalidates all the managed objects of this context which in turn can be an issue if some of these objects are in use. To only reset the changes `context.rollback` needs to be used. 

Please replace `reset` call with `rollback`

```suggestion
				context.rollback()
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,87 @@ public final class CoreDataFeedStore: FeedStore {
...
+					completion(.found(
+						feed: cache.feed
+							.compactMap { ($0 as? ManagedFeedImage) }
+							.map {
+								LocalFeedImage(id: $0.id,
+								               description: $0.imageDescription,
+								               location: $0.location,
+								               url: $0.url)
+							},
+						timestamp: cache.timestamp))
```
danillahtin, line 45: Suggestion:
To improve this part's readability, it can be moved to a separate local constant or a helper method as shown in the course.
```suggestion
					let feed = cache.feed
						.compactMap { ($0 as? ManagedFeedImage) }
						.map {
							LocalFeedImage(id: $0.id,
							               description: $0.imageDescription,
							               location: $0.location,
							               url: $0.url)
						}
					completion(.found(
						feed: feed,
						timestamp: cache.timestamp))
```

danillahtin (commented):
Hi @VSharkO, you almost passed the challenge!

Thank you 🙂

FeedStoreChallenge/CoreDataFeedStore.swift
VSharkO: Hi Danil,
Thank you for the review, it is really helpful :)
When I make ManagedCache and ManagedFeedImage private I get the error: "An NSManagedObject of class 'ManagedCache' must have a valid NSEntityDescription."
I guess this is because of functions in CoreDataHelpers.swift, they don't have access to it...?
So I will move these classes into their own files and make them internal.

VSharkO:
Hi @danillahtin 
Thank you for the review, it was really helpful. 

I made the suggested changes and left one comment here:
https://github.com/essentialdevelopercom/ios-lead-essentials-feed-store-challenge/pull/184#discussion_r694766823

I really appreciate your review. 👍 
Best regards,
Valentin

PR #199, opened 2021-09-09
Joshua Bryson - Feed Store Challenge

DrCarleeknikov:


FeedStoreChallenge/FeedStore.xcdatamodeld/FeedStoreModel.xcdatamodel/contents
```diff
@@ -0,0 +1,18 @@
...
+    <entity name="ManagedCache" representedClassName="ManagedCache" syncable="YES">
+        <attribute name="timestamp" attributeType="Date" usesScalarValueType="NO"/>
+        <relationship name="feed" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="ManagedFeedImage" inverseName="cache" inverseEntity="ManagedFeedImage"/>
+    </entity>
+    <entity name="ManagedFeedImage" representedClassName="ManagedFeedImage" syncable="YES">
+        <attribute name="id" attributeType="UUID" usesScalarValueType="NO"/>
+        <attribute name="imageDescription" optional="YES" attributeType="String"/>
+        <attribute name="location" optional="YES" attributeType="String"/>
+        <attribute name="url" attributeType="URI"/>
+        <relationship name="cache" maxCount="1" deletionRule="Cascade" destinationEntity="ManagedCache" inverseName="feed" inverseEntity="ManagedCache"/>
```
danillahtin, line 12: Deletion rule for `cache` should be `Nullify`. 
If the rule is `Cascade` then the whole `ManagedCache` with all the images will be removed if one of `ManagedFeedImage`s is removed.
Please change the deletion rule to `Nullify`

```suggestion
        <relationship name="cache" maxCount="1" deletionRule="Nullify" destinationEntity="ManagedCache" inverseName="feed" inverseEntity="ManagedCache"/>
```

FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,35 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Joshua Bryson on 9/9/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedFeedImage)
+internal class ManagedFeedImage: NSManagedObject {
```
danillahtin, line 12: `internal` is the default access level, so you don't need to write it explicitly.

It was written explicitly in the first lectures of the program for teaching purposes, but you don't need to do it in your code.

Please, remove all explicit `internal` keywords and make the class `final`.
```suggestion
final class ManagedFeedImage: NSManagedObject {
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,43 @@
...
+//  FeedStoreChallenge
+//
+//  Created by Joshua Bryson on 9/9/21.
+//  Copyright © 2021 Essential Developer. All rights reserved.
+//
+
+import CoreData
+
+@objc(ManagedCache)
+internal class ManagedCache: NSManagedObject {
```
danillahtin, line 12: This class is not intended to be inherited thus it should be `final`
Please make the class `final`.
```suggestion
final class ManagedCache: NSManagedObject {
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,50 @@ public final class CoreDataFeedStore: FeedStore {
...
+			} catch {
+				context.rollback()
+				completion(error)
+			}
+		}
+	}
+
+	private func perform(_ action: @escaping (NSManagedObjectContext) -> Void) {
+		let context = self.context
+		context.perform { action(context) }
```
danillahtin, line 76: Another way of capturing the `context` inside the closure without capturing `self` is to use a capture list as shown below:

```suggestion
		context.perform { [context] in action(context) }
```

It achieves the same result with one less line of code.

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	class func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		let request = ManagedCache.fetchRequest()
+		request.fetchLimit = 1
+		return try context.fetch(request).first as? ManagedCache
+	}
+
+	var localFeed: [LocalFeedImage] {
+		let feed = Array(feed) as! [ManagedFeedImage]
```
danillahtin, line 34: Suggestion:
`NSOrderedSet` can be mapped to array using `compactMap`, it will help to avoid force casting
```suggestion
		let feed = feed.compactMap { $0 as? ManagedFeedImage }
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,43 @@
...
+
+	static func newUniqueInstance(in context: NSManagedObjectContext) throws -> ManagedCache {
+		try deletePreviousCache(in: context)
+		return NSEntityDescription.insertNewObject(forEntityName: String(describing: ManagedCache.self), into: context) as! ManagedCache
+	}
+
+	class func find(in context: NSManagedObjectContext) throws -> ManagedCache? {
+		let request = ManagedCache.fetchRequest()
+		request.fetchLimit = 1
+		return try context.fetch(request).first as? ManagedCache
```
danillahtin, line 30: To slightly increase CoreData performance `request.returnsObjectsAsFaults` can be set to `false`.

```suggestion
		let request = ManagedCache.fetchRequest()
		request.fetchLimit = 1
		request.returnsObjectsAsFaults = false
		return try context.fetch(request).first as? ManagedCache
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,43 @@
...
+}
+
+extension ManagedCache {
+	static func deletePreviousCache(in context: NSManagedObjectContext) throws {
+		try find(in: context).map(context.delete).map(context.save)
+	}
+
+	static func newUniqueInstance(in context: NSManagedObjectContext) throws -> ManagedCache {
+		try deletePreviousCache(in: context)
+		return NSEntityDescription.insertNewObject(forEntityName: String(describing: ManagedCache.self), into: context) as! ManagedCache
```
danillahtin, line 24: Suggestion:
Entity name and class name potentially can be different. Thus it is better to use `entity().name!`
```swift
return NSEntityDescription.insertNewObject(forEntityName: entity().name!, into: context) as! ManagedCache
```
And the whole return can be simplified to the following
```suggestion
		return ManagedCache(context: context)
```

FeedStoreChallenge/ManagedCache.swift
```diff
@@ -0,0 +1,43 @@
...
+@objc(ManagedCache)
+internal class ManagedCache: NSManagedObject {
+	@NSManaged internal var timestamp: Date
+	@NSManaged internal var feed: NSOrderedSet
+}
+
+extension ManagedCache {
+	static func deletePreviousCache(in context: NSManagedObjectContext) throws {
+		try find(in: context).map(context.delete).map(context.save)
+	}
```
danillahtin, line 20: Saving the context violates SRP. This method both deletes the object and saves the context.
Please move `context.save` outside of the function.
```suggestion
	static func deletePreviousCache(in context: NSManagedObjectContext) throws {
		try find(in: context).map(context.delete)
	}
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,50 @@ public final class CoreDataFeedStore: FeedStore {
...
+				completion(.failure(error))
+			}
+		}
 	}
 
 	public func insert(_ feed: [LocalFeedImage], timestamp: Date, completion: @escaping InsertionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				let newCache = try ManagedCache.newUniqueInstance(in: context)
```
danillahtin, line 48: `newUniqueInstance` calls `deletePreviousCache` under the hood which in turn calls `context.save`.
If deletion succeeds and any following operation fails then Core Data will loose the cached instance which is not a good thing. To solve this `context.save` should be moved from `deletePreviousCache` and all the saves should be done only when all the operations succeeded.

danillahtin (commented):
Hi @DrCarleeknikov, you almost passed the challenge!

Thank you 🙂

DrCarleeknikov:
Thanks for the input, I've made those changes!

FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,35 @@
...
+extension ManagedFeedImage {
+	class func images(from images: [LocalFeedImage], in context: NSManagedObjectContext, with cache: ManagedCache) -> NSOrderedSet {
+		var managedFeed = [ManagedFeedImage]()
+		for image in images {
+			let managedImage = ManagedFeedImage(context: context)
+			managedImage.id = image.id
+			managedImage.imageDescription = image.description
+			managedImage.location = image.location
+			managedImage.url = image.url
+			managedImage.cache = cache
```
danillahtin, line 30: No need to set `cache` manually. It will be automatically set when `ManagedFeedImage` is added to `ManagedCache.feed`. Please remove this line
```suggestion
```

FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,35 @@
...
+	@NSManaged var id: UUID
+	@NSManaged var imageDescription: String?
+	@NSManaged var location: String?
+	@NSManaged var url: URL
+
+	@NSManaged var cache: ManagedCache
+}
+
+extension ManagedFeedImage {
+	class func images(from images: [LocalFeedImage], in context: NSManagedObjectContext, with cache: ManagedCache) -> NSOrderedSet {
```
danillahtin, line 22: Suggestion:
Since the class is private, all `class` methods work the same way as `static` ones. The main difference between `static` and `class` (from architectural perspective) is that `class` method can be overridden whilst `static` can't.
Using `class` methods in such a case can be misleading thus it is a little bit more correct to use `static`
```suggestion
	static func images(from images: [LocalFeedImage], in context: NSManagedObjectContext, with cache: ManagedCache) -> NSOrderedSet {
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,50 @@ public final class CoreDataFeedStore: FeedStore {
...
+			}
+		}
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				try ManagedCache.deletePreviousCache(in: context)
+				try context.save()
```
danillahtin, line 66: In case there is no previously stored cache `ManagedCache.deletePreviousCache` will not change the context, thus there will be no changes in it and will be no need to save it. Otherwise, Core Data may perform unnecessary work.
Since there's no information on whether there was actual deletion or not in this scope then `NSManagedObjectContext.hasChanges` property can be used to make a decision.

Please add `hasChanges` check before saving to avoid unnecessary work.
```suggestion
				if context.hasChanges { 
      					try context.save()
      				}
```

danillahtin (commented):
Hi @DrCarleeknikov, almost there!

Thank you 🙂

danillahtin (commented):
Hi @DrCarleeknikov,

✅  All tests are passing 

**[Update your status](https://iosleadessentials.slack.com/archives/CTAR0LVMM/p1629389971188200) in the community now with the following emoji to celebrate your achievement: `:white-belt-2:`**

PR #200, opened 2021-09-10
Nicolò Pasini - Feed Store Challenge

NPasini:


NPasini:
Sorry, I pushed this morning a commit removing the redundant internal access modifier, and this caused the PR to be unvalidated and requiring approval again

mapostolakis:
> Sorry, I pushed this morning a commit removing the redundant internal access modifier, and this caused the PR to be unvalidated and requiring approval again

No problem, @NPasini! The CI is passing and we'll review the challenge shortly 🙂

FeedStoreChallenge/ManagedFeedImage.swift
```diff
@@ -0,0 +1,35 @@
...
+	@NSManaged var id: UUID
+	@NSManaged var imageDescription: String?
+	@NSManaged var location: String?
+	@NSManaged var url: URL
+	@NSManaged var cache: ManagedCache
+}
+
+extension ManagedFeedImage {
+	static func images(from localFeed: [LocalFeedImage], in context: NSManagedObjectContext) -> NSOrderedSet {
+		NSOrderedSet(array: localFeed.map({ local in
```
danillahtin, line 22: Suggestion:
Using implicit `return` is absolutely ok. On the other hand implicit `return`s are not used in this project thus it is better to be consistent and follow the project's code style
```suggestion
		return NSOrderedSet(array: localFeed.map({ local in
```

FeedStoreChallenge/CoreDataFeedStore.swift
```diff
@@ -29,14 +29,50 @@ public final class CoreDataFeedStore: FeedStore {
...
+		}
 	}
 
 	public func deleteCachedFeed(completion: @escaping DeletionCompletion) {
-		fatalError("Must be implemented")
+		perform { context in
+			do {
+				let managedCache = try ManagedCache.newUniqueInstance(in: context)
+				context.delete(managedCache)
+				try context.save()
```
danillahtin, line 66: `ManagedCache.newUniqueInstance` creates a new instance which is extra work and is not required by this method. `ManagedCache.find` fits better.
In case there is no previously stored cache `context.delete` will not change the context, thus there will be no changes in it and will be no need to save it. Otherwise, Core Data may perform unnecessary work.
```suggestion
				if let managedCache = try ManagedCache.find(in: context) {
					context.delete(managedCache)
					try context.save()
				}
```

Another way to ensure the context is only saved when there are changes is to use `.map` instead of an `if` statement:

```suggestion
				try ManagedCache.find(in: context)
					.map(context.delete)
					.map(context.save)
```

danillahtin (commented):
Hi @NPasini, you almost passed the challenge!

Thank you 🙂

NPasini:
@danillahtin Thank you for the suggestions, I have applied them and pushed the changes

danillahtin (commented):
Hi @NPasini,

✅  All tests are passing 

**[Update your status](https://iosleadessentials.slack.com/archives/CTAR0LVMM/p1629389971188200) in the community now with the following emoji to celebrate your achievement: `:white-belt-2:`**

